PR #21, opened 2017-01-07
Update discord.js

dillman33:


SplitPixl:
That's not how this works...

Ratismal:
I don't know if you're joking or not, but just in case
1. I have no intention on ever using discord.js
2. Simply changing the import would not port the bot to discord.js, making this a highly breaking change
3. All eris functions would have to be rewritten using their discord.js counterparts
4. If you actually want me to switch to discord.js, then have fun porting all my code for me. I still wouldn't merge if it executed perfectly.

hydrabolt:
:(

Ratismal:
@hydrabolt it's not that I don't like discord.js. There are many aspects of it that I prefer over eris. I just don't have the time or need to port all of my code to it.

hydrabolt:
It's fine hahaha I was just kidding :P

PR #22, opened 2017-01-07
Added increment and aincrement tags

Titansmasher:


Ratismal:
I'm not sure if I want two tags for every variable manipulation operation. I'll keep this PR open, but I'm still brainstorming better methods (only reason I haven't implemented myself)

PR #26, opened 2017-01-25
added Date tag

Vap0r1ze:
You need to do `npm i moment-timezone --save` before adding this


Ratismal:
I'll add this functionality to the existing {time} tag.

Don't worry, I already have moment-timezone. All the moment libs are my secret fetish.

PR #29, opened 2017-01-27
escape link in help message

bippum:


Ratismal:
Implemented, sorry for the delay.

PR #43, opened 2017-10-13
Quick fix to issue with numbers in if

Titansmasher:
Added a quick fix to includes/startswith/endswith that was causing issues with execution when pure numbers were supplied to the if tag


Ratismal:
sorry it took so long, the cast is only neccesary on the `a` variable and needs to be applied to the {bool} tag as well. I've implemented this myself.

Thanks though!

PR #50, opened 2018-01-24
Update tags.js

Titansmasher:
Im not sure why this exists, however adding a zws to all tags containing the word "help" anywhere inside them seems a little strange.
Removing this hopefully will fix some unexpected outcomes


Titansmasher:
Closing because there is actually some reasoning behind having this there

PR #57, opened 2018-03-07
Updated the copyright year

byparallax:
2016 -> 2018


Ratismal:
it would still be broken on every other page. i will get around to doing this eventually.

PR #59, opened 2018-03-17
Fixes for greet and farewell

Titansmasher:


Ratismal:
would it be possible for you to resubmit this PR, with less commits? :eyes: i wanna see what the problem actually was

Titansmasher:
Im not sure how to do that ;-; The commits are in my master branch now. 
You should be able to look at the [files changed](https://github.com/Ratismal/blargbot/pull/59/files) page to see that right?

Ratismal:
Oh, that's a thing. I was just clicking on the commits which gave me every single change that was made and unmade. Thanks.

Titansmasher:
It might make sense for you to just duplicate the changes I made (if you accept the fix) and do it yourself, so as to not clutter the commit history

Ratismal:
Commit history is fine as it will be squashed. However, I can't remember what this was supposed to *fix*, and thus what I should be testing for. Do you remember?

Titansmasher:
greet and farewell were failing to send at all if you didnt have a -c parameter on the command.
I made it default to the current channel

Ratismal:
:ok_hand:

PR #60, opened 2018-03-18
Bbtag update phase1

Titansmasher:
Ive done a bunch of work on all of the subtags. The majority of it falls into these categories:
1. **Migration to a new tag system.** A lot of common functionality has been extracted into TagBuilder and thats now used instead of defining a new object from scratch
2. **Enforcing upper limits on argument counts.** All subtags that do not have a greedy parameter now throw an error if you provide too many arguments. Ive done this because leaving it open ended means people can easily misunderstand what the subtag is doing and insert extra, unused values that will reduce maintainability
3. **Added a few new subtags.** Namely {embed}, {addreact}, {buildembed}, {quiet}, {zws}, {iscc}, {color}
4. **Extending functionality.** I have added a bunch of extra functionality to existing subtags, hopefully without breaking any existing functionality. e.g. {nsfw} now must be run for it to mark the output as NSFW, and can supply a custom error message. {return} now has the option to stop at an encasing {exec}

There is now also an additional subcommand for CCommand and Tag, which will show you the docs for any subtag. This respects the DMHelp setting, and can be accessed via `b!cc docs <subtag>` and `b!t docs <subtag>`

I do have plans for more changes (hence this being phase 1) but due to the scale of this PR already, I figured it would be best to leave them to phase 2 and 3. Some of the changes I am planning to include are:

Phase 2 (Engine):
1. Rewrite the engine behind the BBTag language to process everything in 2 passes (ignoring exec, execcc and inject, they will need an additional 2 passes per instance) As well as enabling better error reporting. Errors will be able to be displayed with a column and row location for ease of debugging
2. Add a debug subcommand to tag and cc, which will allow you to view all working variables and errors thrown by a tag after completion. Should be very useful for additional debugging

Phase 3 (New stuff):
1. Add a bunch of new subtags that I kept being pestered to add into phase 1. Phase 1 only includes some of the more flashy ones, to keep people happy, but the majority of new subtags will be here.


byparallax:
image

Ratismal:
time to watch the world burn

PR #76, opened 2018-03-31
Fix comments

Titansmasher:
Titan is a moron


Ratismal:
does resolveArgs need to be before whenDefault?

Titansmasher:
I designed this all around a "fluent" api kinda style. Things can be done in any order before the .Build()
I just use the order I do for consistency between subtags in how they look

Ratismal:
well, gonna close the PR since I've already fixed this

PR #93, opened 2018-05-07
Update and rename isstaff.js to ismod.js

HunteRoi:
Consistency change (because of `b!mods`)


Titansmasher:
I feel this should be left as isstaff, with ismod as an alias.
`b!mods` is the only place its referred to as mod, in the settings, in the subtag documentation, and in the code it is called staff in every other instance.

PR #94, opened 2018-05-12
exec/execcc patch

Zoomah:
Allows to seperate arguments for {exec} and {execcc} with semicolon, like this:

{exec;myTag;argument1;argument2;...}


src/tags/exec.js
```diff
@@ -1,4 +1,5 @@
 /*
+import { function } from './../backend/public/codemirror/mode/shell/shell';
```
Ratismal: 👀

Zoomah: I didn't do THAT that was there

PR #105, opened 2018-06-08
Small important fix?

HunteRoi:
Hey there!

This pull requests contains two changes : 
  - a**n ugly** fix to prevent people from toggling off the `editcommand`. I also added a line about `cannotDisable` property in the [BaseCommands.js file](https://github.com/HunteRoi/blargbot/blob/302d1fa4159f9b9f98beb7de1431ca05b2390edc/src/structures/BaseCommand.js) but it's commented as I am not sure it has its place there. It's been repercuted to the `editcommand` file as well (but still as comments). 
Mention me on Discord or answer this thread so I can understand how I can fix issues like this in the future please! 😄 

  - a **possibly breaking** fix for the __**Flags:**__ bit appearing in every `help` message for default commands. "Possibly breaking" in the way that I have no idea if you check everywhere in your code that the `flags` property is undefined or not...


HunteRoi:
Included in Titan's PR.

PR #112, opened 2018-06-15
Updates for V6.10

Titansmasher:
## Features & Improvements
- [x] Improve emoji parsing & rendering
  - Use the twemoji library to find emoji rather than emoji-regex
- [x] `b!emoji` use local svg's rather than remote
- [x] Display line & column rather than character position in errors
- [x] Add `{output;<text>}`
  - If specified, this replaces the default output of tags, allowing for code to be run after sending in tags
- [x] Add `{lock;<mode>;<key>;<code>}`
  - `<mode>` can be `read` or `write`. reads can run in parallel while write must run alone
  - `{lock}` cannot nest any locks or waiting subtags
- [x] Enable `{edit}` in a similar fashion to `{delete}`
- [x] Add `{function;<name>;<code>}`
  - Creates a new temporary subtag that may be used, effectively creating custom functions
  - Function names are auto-prefixed with `func.`
  - Parameters are provided through `{params;[index];count}`, `{paramsarray}` and `{paramslength}` temporary subtags
- [x] Turn help messages into embeds. If a channel cannot accept embeds, return as a string as normal
- [x] Allow users to be ignored when it comes to the modlog
  - `b!log ignore` to add `b!log track` to unignore
- [x] Add timer management on a per guild basis
  - Only timers created after this update will appear on this list due to data not existing in prior versions of the bot
- [x] Users can now be excluded from the event log (`b!log <ignore | track> <users...>`)

## Bugs
- [x] Fix bbengine not detecting a missing brace when a child also closes
  - `{subtag;{child}` should be classed as invalid
- [x] Fix [undefined in help message](https://get.banned.today/ecdb28.png)
- [x] Various security/performance fixes
  - `{roleadd}`, `{roleremove}`, `{rolecreate}`, `{roledelete}` and `{rolesetmentionable}` now check the tag authors permissions.
  - Add limits to various discord functions
    - `{send}` limited to 10/tag
    - `{delete}` limited to 11/tag
    - `{edit}` limited to 10/tag
  - Queries for users and roles are now limited to 5 per tag. All queries performed after this will have `quiet` and `suppress` forcibly set to true
  - Queries now cache the result for each input. e.g. `{userid;a}{userid;a}` will only display 1 query prompt
  - Arguments for subtags are now limited to 1,000,000 characters long by default
    - The regex subtags have been limited to 50,000 characters and 2,000 characters for the actual regex
- [x] Fix `{edit}` cannot remove embeds or text from message
  - If the message or embed is set to `_delete` then the text/embed will be removed from a message
- [x] Fix severe performance issues with `{exec}` and `{execcc}`

## Cleanup
- [x] Restructure BBTagEngine to be more modular


Ratismal:
package.json conflict is a mess. i added the package `big-integer`, what did you add? it's just hard to tell because diff is getting upset about the indent change. let me get into individual feedback. if a point isn't listed, that means im fine with it

## Features & Improvements
- Add `{output;<text>}`
  - what exactly is the behaviour of this? could i get an example?
- Add `{lock;<mode>;<key>;<code>}`
  - `{lock}` cannot nest any locks or waiting subtags
    - does this need to even execute nested subtags?
- Enable `{edit}` in a similar fashion to `{delete}`
  - is this referring to the ability to use {edit} in tags, as long as they apply to related messages
- Add `{subtag;<name>;<code>}`
  - im assuming that this is the {function} pr, will put my notes in there
- Disable logging of the bots own messages
  - is this referring to chatlogs or event logs? because i'd be fine with this for the latter, but not the former

## Bugs
- `{addrole}` & `{removerole}` oversight
  - what exactly was this oversight?

## Cleanup
- Restructure utils (`bu`) to be more modular
- Remove `dep`
  - all i can say for these two points is "good luck". let me know if you want me to help out with this to ease the workload

Ratismal:
here we go

PR #117, opened 2018-06-16
Add {function} subtag

Titansmasher:


Ratismal:
assuming that this is the {subtag} subtag referred to in #112 

only feedback would be that functions would be called by {func.name}, as discussed in dms

PR #133, opened 2018-07-15
Update ready.js

HunteRoi:
Fixing the problem where the bot changes its status several time in a second.


Saederup92:
I guess you can say Blarg is in a complicated relationship

PR #134, opened 2018-07-16
 Fixed params overrides not being reverted

SarahShmarah:


Titansmasher:
Looks good to me. Tests all looked fine.
@Ratismal I think this is ready for a merge

PR #138, opened 2018-07-19
Fix description

loliprotector:


Ratismal:
eugh i thought i fixed this

PR #140, opened 2018-07-19
Role-restricted emojis

HunteRoi:
**Known fact :** you can add emojis to your server through the API.
**Unknown fact until now :** you can add role restriction to emojis.

```js
msg.guild.emojis.create('LINK', 'NAME', { roles: [ 'ROLEID' ], reason: 'Log in audits' })
```
If the bot doesn't have the role to use the emoji, it encounters a problem returning an array of emojis. The reason is still unknown. Maybe that API-side, the bot doesn't get the emoji ID as it doesn't have the role to access it.

Two fixes :
- first and easy fix is just to ignore role-restricted emojis.
- second one is to check if the bot has the roles to see the restricted emojis and add the emoji to the list if it does.


HunteRoi:
@Ratismal Second solution seems better to me as it's completely standardized to "one syntax".
But @byparallax's will is attractive too. It's kinda part of what I've written there : a possibility of ignoring non-accessed emojis.
I'm for both ¯\_(ツ)_/¯

Titansmasher:
Personally I disagree with having the subtag return without the <>'s ever, because thats then not a valid emoji in discords eyes. I like parallaxes suggestion to have a control arg to filter down the output.
```{emojis}``` => ```["<:public:11111>","<a:rolerestricted:22222>"]```
```{emojis;roleid}``` => ```["<a:rolerestricted:22222>"]```

I understand the rationale behind showing without the <> however that then is no longer considered an emoji by discord. Should people be forced to transform the output before its usable?

PR #150, opened 2018-08-09
Implement MD5 tag

byparallax:
This is either broken or the worst way to do it. No idea, first time writing javascript. 
/shrug


byparallax:
Also this is a vanilla implementation because I have no idea how to do stuff 🤔 
I guess I can try a version using the "md5" npm module, but then it requires doing `npm install md5` so ehh

byparallax:
Yay! Now it's using the package [md5](https://www.npmjs.com/package/md5). 
So please be kind enough to run `npm install md5`. 

Oh, and I guess it's probably still broken.

PR #162, opened 2018-09-12
Tweak unsafe regex detection and mention it in the docs

### LINKED ISSUE #159: Consider tweaking evil regex detection
In my humble opinion, the current state of evil regex detection may cripple useful functionality of the Javascript regular expression engine to detect certain complex (but common) patterns, without necessarily compromising that functionality for decreased resource usage of the bot on its servers.

For instance, the following Javascript regex for matching e-mail addresses according to RFC 5322 (taken from http://emailregex.com/) fails to compile on Blargbot because it is considered evil:

`{regexmatch;myemail@server.com;/^(([^<>()\[\]\\.,{semi}:\s@"]+(\.[^<>()\[\]\\.,{semi}:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/}`

Another regex that failed to compile on Blargbot because this reason was [this one I came up with to validate URIs according to RFC 3986](https://regex101.com/r/jA6jJ7/5) (it might indeed be evil, but I really hope it is not. Others I found on Internet can even get more complex, because they validate the authority field too and are even more standards-conforming). As you can test on the interactive Javascript regex page I linked to, its performance seems _good enough_ for arbitrary input strings (I've tested it with several pretty long Google Image result data URIs and the matches were returned in less than 3 ms). For Blargbot I used a shortened version that recognizes less schemes, in order to stay below the 2000 character limit, without any more changes.

It's probable you're wondering why I want such complex regex matching for a Discord server. The answer is that currently I'm coding an everything autoresponse in BBTag to give users warnings automatically if they use caps so much, and I wish to be able to ignore their message if it is just an URI, because it makes little sense to warn users for caps used in a link path or query strings (the e-mail regex was just another example I thought it would be worth mentioning; I have no plans on matching e-mail addresses right now). Of course, I probably could have done an easier (but more error-prone) regex that works only with common URI formats as those used on HTTP(S), and it would probably be _good enough_, but stumbling upon this gave me an improvement suggestion for Blargbot anyway (_good enough_ doesn't mean _not able to be improved_).

I came up with serveral proposals to address this issue (you might find them unapplicable or simply plain bad ideas, though):

- Clearly warn in the documentation that evil regexes are checked for and won't compile, and hint what kind of regexes could be considered evil. _This doesn't fix the issue itself, but at least it would stop programmers like me spending energy to build a expectedly-very-good regex to only find it fail on Blargbot._

- Consider removing [this apparently spurious check](https://github.com/blargbot/blargbot/blob/362896dd83b922d9023a35ec7fdc063118ffbbc1/src/utils/generic.js#L1385) that marks as unsafe any regex which contains a capture group with the `+` or `*` qualifiers inside, being the capture group appended by the `+` or `*` qualifiers. _At best, I think that the usage of the `safe-regex` library that follows should be enough to detect evil regexes with _good enough_ accuracy and false positives. And this check doesn't take into account alternative qualifiers with similar effects such as `{1,}`. A library should be able to detect these trivial cases of evil regexes better, with less false positives, and if not then I'd consider switching to other library._

- Study the applicability of more efficient ways to compile and execute regexes, so evil regexes could be dealt with in a more graceful manner than simply refusing to compile them. _For instance, the regex matching algorithm could be run in a separate Javascript thread, which is aborted if it takes too long to run. Or improve the regex engine itself, by ensuring it runs in O(n) time for the most part (this is possible at least theoretically, I don't know if it is feasible to implement on Javascript)._

Some interesting reference I've found about the subject:

https://en.wikipedia.org/wiki/ReDoS#Description
https://www.reddit.com/r/programming/comments/gua22/regex_denial_of_service_redos/
https://swtch.com/~rsc/regexp/regexp1.html

I hope that was not too long to read 😄

AlexTMjugador:
This PR proposes the removal of an apparently spurious unsafe regex check, that should be already covered by the library call which follows with less false positives, and updates the documentation of the commands and tags which use `bu.createRegExp` to indicate that regexes will be checked against safety and length.

I did this PR in a bit of a rush from my web browser. I think it should work, as it is a simple check removal and string modification, but perhaps I missed something or I did something essentially wrong, so I'm open to changes.

Resolves #159.

### DIFF
--- a/src/dcommands/autoresponse.js
+++ b/src/dcommands/autoresponse.js
@@ -11,7 +11,7 @@ class autoresponseCommand extends BaseCommand {
             flags: [{
                 flag: 'R',
                 word: 'regex',
-                desc: 'If specified, parse as /regex/ rather than plaintext.'
+                desc: 'If specified, parse as /regex/ rather than plaintext. Unsafe and very long (more than 2000 characters) regexes will not parse successfully.'
             }, {
                 flag: 'e',
                 word: 'everything',
--- a/src/dcommands/censor.js
+++ b/src/dcommands/censor.js
@@ -10,7 +10,7 @@ class CensorCommand extends BaseCommand {
             flags: [{
                 flag: 'R',
                 word: 'regex',
-                desc: 'Add/Edit: If specified, parse as /regex/ rather than plaintext.'
+                desc: 'Add/Edit: If specified, parse as /regex/ rather than plaintext. Unsafe and very long (more than 2000 characters) regexes will not parse successfully.'
             },
             {
                 flag: 'w',
--- a/src/tags/regexmatch.js
+++ b/src/tags/regexmatch.js
@@ -13,7 +13,9 @@ module.exports =
     Builder.ArrayTag('regexmatch')
         .withAlias('match')
         .withArgs(a => [a.require('text'), a.require('regex')])
-        .withDesc('Returns an array of everything in `text` that matches `regex`.')
+        .withDesc('Returns an array of everything in `text` that matches `regex`. ' +
+            '`regex` will only succeed to compile if it is deemed a safe regular expression ' +
+            '(safe regexes do not run in exponential time for any input) and is less than 2000 characters long.')
         .withExample(
             '{regexmatch;I have $1 and 25 cents;/\\d+/g}',
             '["1", "25"]'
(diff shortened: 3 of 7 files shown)


AlexTMjugador:
Hello. I'd like to know if there are any plans to merge this PR or reject it, after half a month since I've submitted this PR. I kindly remind that I can refactor or modify my changes if needed. Thank you.

Ratismal:
Whoops, I forgot about this. Looks good to me!

PR #166, opened 2018-09-28
Fix docs and subcommand help message display

HunteRoi:
- Fixing the documentations at some place
- Fixing an issue with the subcommand help messages not being displayed

### DIFF
--- a/src/dcommands/tag.js
+++ b/src/dcommands/tag.js
@@ -202,7 +202,7 @@ class TagCommand extends BaseCommand {
                 + '\n__**Subcommands:**__\n'
                 + `${subcommands.map(x => `**${x.name}**`).join(', ')}`
                 + '\nFor more information about a subcommand, do `b!tag help <subcommand>`.\n'
-                + '\nFor more information about BBTag, visit <<https://blargbot.xyz/tags\>.\n'
+                + '\nFor more information about BBTag, visit \<https://blargbot.xyz/tags\>.\n'
                 + 'By creating a tag, you acknowledge that you agree to the Terms of Service (\<https://blargbot.xyz/tags/tos\>)'
         });
     }


src/dcommands/tag.js
```diff
@@ -433,7 +433,7 @@ class TagCommand extends BaseCommand {
                         });
                         if (command.length > 0) {
                             await bu.send(msg, `Subcommand: **${command[0].name}**
-Aliases: **${command[0].aliases.join('**, **')}**
+Aliases: **${command[0].aliases ? command.aliases.join("**, **") : "none"}**
```
Ratismal: `.aliases` is not a property of `command`. change this to `(command[0].aliases || []).join('**, **') || 'none'`

src/dcommands/ccommand.js
```diff
@@ -124,8 +124,8 @@ class CcommandCommand extends BaseCommand {
                 + 'For more in-depth command customization, see the `editcommand` command.\n'
                 + '\n**Subcommands:**\n'
                 + `${subcommands.map(x => `**${x.name}**`).join(', ')}`
-                + '\nFor more information about a subcommand, do `b!cc help <subcommand>.`\n'
-                + '\nFor more information about BBTag, visit <https://blargbot.xyz/tags>.'
+                + '\nFor more information about a subcommand, do `b!cc help <subcommand>`.\n'
+                + '\nFor more information about BBTag, visit https://blargbot.xyz/tags.'
```
Ratismal: i'd much rather this stay escaped

src/dcommands/ccommand.js
```diff
@@ -655,7 +655,7 @@ class CcommandCommand extends BaseCommand {
                         });
                         if (command.length > 0) {
                             bu.send(msg, `Subcommand: **${command[0].name}**
-Aliases: **${command[0].aliases.join('**, **')}**
+Aliases: **${command[0].aliases ? command.aliases.join("**, **") : "none"}**
```
Ratismal: `.aliases` is not a property of `command`. change this to `(command[0].aliases || []).join('**, **') || 'none'`

src/dcommands/tag.js
HunteRoi: I've forgotten an [0] in there!

Although, I'm not sure why you want 
`Aliases: **${(command[0].aliases || []).join("**, **") || 'none'}` instead of `Aliases: **${command[0].aliases ? command[0].aliases.join("**, **") : "none"}**`

src/dcommands/ccommand.js
HunteRoi: Previously, the content was `<https://...>` which is currently displayed like this on the website :
image

Wanted to get rid of that problem so I removed the characters that were recognized by the parser.
I'm unsure if `\<https://...\>` would work or not but it's still the thing I've changed for your request.

src/dcommands/ccommand.js
HunteRoi: I've forgotten an [0] in there!

Although, I'm not sure why you want 
`Aliases: **${(command[0].aliases || []).join("**, **") || 'none'}` 

instead of 

`Aliases: **${command[0].aliases ? command[0].aliases.join("**, **") : "none"}**`

PR #167, opened 2018-10-01
Remove econ.js command

byparallax:
fixer.io now requires a paid plan in order to convert currencies - until fixed, it is better to delete it in order to avoid disappointment.


Ratismal:
agreed, will disable

PR #172, opened 2019-02-02
Update reactlist.js

HunteRoi:
Add a use of the pagination system to get all the users when the subtag is supposed to return the userids array


HunteRoi:
I'm thinking about the `null` value in the call <https://github.com/HunteRoi/blargbot/blob/patch-1/src/tags/reactlist.js#L70>. It could be better to use `undefined` instead... I really don't know.
This new implementation might need some testing !!

PR #179, opened 2019-06-21
Bugs/splitinput newline

danny-may:
Reworked the splitinput method to respect all whitespace characters, not just spaces.
Reworked to use a generator function to try to make it more readable.

Sample test:
image

### DIFF
--- a/src/utils/generic.js
+++ b/src/utils/generic.js
@@ -960,9 +960,14 @@ bu.splitInput = (content, noTrim) => {
     return [...splitInput(content, noTrim)]
 };
 
+/**
+ * Produces a series of strings from the given content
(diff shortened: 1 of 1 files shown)


Ratismal:
Sorry it's taken so long to get back to you on this. My primary concern is if this breaks commands that should retain newlines, such as `b!t set <name> <value>`. This is the reason I've had so many issues with splitInput in the past - there's so many cases where it needs to be handled differently, and we don't have a way to account for that.

Possible solutions I can think of are
1. add another flag (similar to notrim) which would prevent newlines from being parsed. very messy.
2. abstract input parsers to a sort of strategy pattern, where commands can specify which pattern they utilize

PR #180, opened 2019-07-05
Bugs/reactlist

danny-may:


src/tags/reactlist.js
```diff
@@ -59,22 +61,27 @@ module.exports =
             // List all users per reaction
             let users = [];
             let errors = [];
-            for (const emote of parsed)
+            for (const emote of parsed) {
```
Ratismal: reassignment of const `emote` will break
```suggestion
            for (let emote of parsed) {
```

Titansmasher:
Whoopsie

PR #181, opened 2019-07-12
Fix the analysis of subtags at the point of setting

danny-may:
I know this PR is a bit rough, but I have plans for a proper implementation in the rewrite I am doing.

Changes from the current implementation:

1. No analysis within `{//}`
2. `{params}`, `{paramsarray}` and `{paramslength}` no longer report as errors when used within `{function}`
3. `{lock}` reports an error when used within `{lock}`
4. Restrictions on `{waitreact}`, `{waitmessage}` and `{filter}` are reported

### DIFF
--- a/src/core/bbtag.js
+++ b/src/core/bbtag.js
@@ -408,6 +408,10 @@ function* analyzeSubtag(subtag, parents) {
         return;
     } else if (name.children.length > 0) {
         yield { subtag, warning: 'Dynamic subtag' };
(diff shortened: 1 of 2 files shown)


HunteRoi:
@danny-may should this be closed or merged ?

HunteRoi:
@danny-may :up:

danny-may:
Ew, my old code  burn it in fire

PR #190, opened 2020-02-15
List custom commands by roles

### LINKED ISSUE #189: No way to check what roles can access what cc
There is a `setrole` subcommand for custom commands but there is no way to check what roles can use which commands.
I consider this an important feature. Why would we have a `set` without its `get` ?

HunteRoi:
Fixes #189


Saederup92:
Should this be useable by anyone in a guild? 🤔

HunteRoi:
> Should this be useable by anyone in a guild? 🤔

The changes are made on the `ccommand` built-in command which is by default restricted in use to `Admin`. So the answer is no. 
It's more a matter of "display more information" for the management of custom commands instead of a guild friendly change bound for anyone.

PR #192, opened 2020-03-07
Add subcommand "setauthorizer" to claim authorizer role of a ccommand

HunteRoi:
Add a way for cc's owner to change the authorizer of the custom command with just a subcommand.

- [x] Update the `bu.ccommand` object to include a `setauthorizer()` method
- [x] Update the `bu.ccommand` object to include a `getauthorizer()` method
- [x] Update the `ccommand` to include a subcommand in order to change the authorizer

### DIFF
--- a/src/utils/database.js
+++ b/src/utils/database.js
@@ -132,6 +132,21 @@ bu.ccommand = {
         r.table('guild').get(guildid).replace(storedGuild).run();
         return;
     },
+    setauthorizer: async function (guildid, key, authorizer) {
+        let storedGuild = await bu.getGuild(guildid);
+        key = key.toLowerCase();
+        if (!storedGuild || !storedGuild.ccommands[key]) return false;
+        
+        storedGuild.ccomands[key].authorizer = authorizer;
+        r.table('guild').get(guildid).replace(storedGuild).run();
+        return true;
+    },
+    getauthorizer: async function (guildid, key) {
+        let storedGuild = await bu.getGuild(guildid);
+
+        if (!storedGuild || !storedGuild.ccommands[key.toLowerCase()]) return undefined;
+        return storedGuild.ccommands[key.toLowerCase()].authorizer;
+    },
     sethelp: async function (guildid, key, help) {
         let storedGuild = await bu.getGuild(guildid);
 
--- a/src/utils/database.js
+++ b/src/utils/database.js
@@ -137,7 +137,7 @@ bu.ccommand = {
         key = key.toLowerCase();
         if (!storedGuild || !storedGuild.ccommands[key]) return false;
         
-        storedGuild.ccomands[key].authorizer = authorizer;
+        storedGuild.ccommands[key].authorizer = authorizer;
         r.table('guild').get(guildid).replace(storedGuild).run();
         return true;
     },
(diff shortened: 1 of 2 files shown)


Ratismal:
So, the author check doesn't seem to be working for me.
image

However, my main concern is what the _purpose_ of this change is.

So first of all, the `authorizer` of a command is basically how the bot tells what permissions the command should have, in an attempt to prevent potential abuse. So for example, if someone with no permissions creates and authorizes a custom command that tries to ban everyone, it'll fail since the command hasn't been authorized to do so.

So my concern would be, in that situation, the user could simply authorize the owner of the guild and ban everyone.

A different solution could be that `setauthorizer` sets the authorizer to only the user executing the command. This ensures that users cannot exploit by assigning the authorizer to a higher up. It also removes the issue for if the author completely leaves the guild.

What exactly was the use case that prompted this change?

HunteRoi:
Let's say you set up a command within a guild you are not really planning to stay as Staff. You're the owner and authorizer of the command.
If you leave or decide to give up your rank, you still stay the author and authorizer. The command, if it uses admin subtags, may not work anymore and forces the Staff to look at it, maybe copy-paste everything again to "get the ownership". Or, else : connect to the IDE and save the command to claim the ownership as well.

To deal with it, I thought about a quick way to just give someone the authorizer role. However, you are right : I did not think about the case where owner could make a malicious command and grant the authorizer role to someone higher in the hierarchy. It didn't cross my mind because `ccommand` is built as an admin command by default.

HunteRoi:
@stupidcat Would it make sense to remove the `[authorizer]` and use the guild member executing the command, and rely on the ccommand default permissions for protection ? 

So in that case, this will follow your previous comment and
- remove the issue where the owner leaves
- remove the concern about user assigning owner as cc authorizer
- let the guild decide if it's safe or not to give `ccommand` to lower role(s) than Admin (which is by default)

Also, the problem with detecting the owner of a tag is my fault obviously. I was too tired and mistook `author = await r.table('user').get(tag.author).run();`. I thought it would return the author object but it only returns the id. 
I will drop my commit #ae07954 and change `author` to `authorId` so it's understandable.

HunteRoi:
Ok found out that I should use `author.userid` instead of `author.id`.

RagingLink:
Any idea when this might be merged? @Ratismal

HunteRoi:
:up:

RagingLink:
LGTM, will test this to make sure everything works

src/utils/database.js
```diff
@@ -132,6 +132,21 @@ bu.ccommand = {
         r.table('guild').get(guildid).replace(storedGuild).run();
         return;
     },
+    setauthorizer: async function (guildid, key, authorizer) {
+        let storedGuild = await bu.getGuild(guildid);
+        key = key.toLowerCase();
+        if (!storedGuild || !storedGuild.ccommands[key]) return false;
+        
+        storedGuild.ccomands[key].authorizer = authorizer;
```
RagingLink, line 140: Looks like there is a small typo here! Everything else seems to work fine 👍

RagingLink (changes requested):

RagingLink (approved):

PR #198, opened 2020-04-14
Handle provided empty condition

### LINKED ISSUE #199: waitreact and waitmessage condition not defaulting to true
As stated by **RagingLink#9469** via the bug-report system, the following subtags have an optional parameter `condition` which should default to true if it isn't provided :
- `{waitmessage;[channels];[users];[condition];[timeout]}`
- `{waitreaction;<messages>;[users];[reactions];[condition];[timeout]}`

Subtags return `Condition must return 'true' or 'false'` with use case where it actually is an empty string and so shouldn't be interpreted as provided value : `{waitmessage;{channelid};{userid};;60}`.

Feedback available [here](https://blargbot.xyz/feedback/rec5myzRJj8Did5sx)

HunteRoi:
Fixes #199 

Update the check statement for the `condition` optional argument to include the case where it could be an empty string within:

- [x] `waitmessage.js` subtag where the BBTag code is stored in `args[2]` ([7851b71](https://github.com/blargbot/blargbot/commit/7851b7130a1620e446735c647e5b2df9baf62643))
- [x] `waitreact.js` subtag where the BBTag code is stored in `args[3]`  ([7851b71](https://github.com/blargbot/blargbot/commit/7851b7130a1620e446735c647e5b2df9baf62643))

### DIFF
--- a/src/tags/waitmessage.js
+++ b/src/tags/waitmessage.js
@@ -92,7 +92,7 @@ module.exports =
             }
 
             // parse check code
-            if (args[2]) {
+            if (args[2] && args[2] !== "") {
                 checkBBTag = args[2];
             } else {
                 checkBBTag = bbengine.parse("true").bbtag;
--- a/src/tags/waitreact.js
+++ b/src/tags/waitreact.js
@@ -74,7 +74,7 @@ module.exports =
             }
 
             // parse check code
-            if (args[3]) {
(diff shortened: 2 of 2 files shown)


HunteRoi:
Another shorter way would be to evaluate `Boolean(args[x])` instead of `args[x] && args[x] !== ""`

PR #201, opened 2020-05-15
Add prefix to ccommands and tags when added to the context cache to prevent conflicts

HunteRoi:
- Edit `{exec}` subtag to add `tag_` prefix when adding a public tag to the context cache ([f895bfd](https://github.com/blargbot/blargbot/commit/f895bfdd63a678b37c6ad883152d3f544a67245d))
- Edit `{execcc}` subtag to add `cc_` prefix when adding a ccommand to the context cache ([1b0abe7](https://github.com/blargbot/blargbot/commit/1b0abe7beef07b833b1ff3cdbac33163c5139d70))

Both to prevent conflicts when custom commands and public tags are using each other and encounter a `getCached` conflict.

An example of this issue would be the **colorme** cc calling the **colourme** cc which contains `{exec;colourme;{args}}`. What happens when **colorme** is executed:
- it adds the **colourme** ccommand to the context's cache,
- it checks for the **colourme** cached value, finds one (which is the ccommand, not the tag which still needs to be added but will not),
- it executes the usertag with the provided context not having the public tag in it.

image

### Thanks to RagingLink#9469 for the issue breakdown and research.

### DIFF
--- a/src/tags/exec.js
+++ b/src/tags/exec.js
@@ -20,7 +20,7 @@ module.exports =
         )
         .whenArgs(0, Builder.errors.notEnoughArguments)
         .whenDefault(async function (subtag, context, args) {
-            let tag = await context.getCached(args[0], key => r.table('tag').get(key).run());
+            let tag = await context.getCached(`tag_${args[0].toLowerCase()}`, key => r.table('tag').get(key).run());
 
             if (tag == null)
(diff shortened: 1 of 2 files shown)


HunteRoi:
@Ratismal any update for the review?

RagingLink (approved):
👍🏻

PR #204, opened 2020-05-26
Added logic for empty condition in waitreact and waitmessage

### LINKED ISSUE #199: waitreact and waitmessage condition not defaulting to true

RagingLink:
Fixes #199 
Previously when `[condition]` was provided an empty string it would return `Condition must return 'true' or 'false'` while it should default to true.
This is because the empty string is parsed with BBtag.parse('') and thus returns a populated Object from the `BBtag` Class. By calling `.content` on that Object and checking if it's not an empty string it will go to the `else` of the `if` statement and default to `true`

**Added:**
- Check for args[x].content if it's an empty string


HunteRoi:
Hey @RagingLink ! Thank you for your insight on issue #199 and this PR that will fix the default behavior. Can you please start checking at the conflict before @Ratismal takes the time to look at your code suggestion?

Have a Merry Christmas and Happy New Year ! Wish you all the best for 2021 🎉 🎅

HunteRoi (approved):

RagingLink:
@HunteRoi thanks for pointing it out! Should be resolved now (I'm still very new to Git and Github so I hope it's all good).

Merry Christmas and Happy New Year to you too! 🎊

PR #206, opened 2020-06-02
Fixed .hasPerm() not working with roleID

### LINKED ISSUE #205: Unclear that custom role requirement needs the rolename as a string
Command: b!editcommand setrole

By the documentation it's impossible to tell what format the role in the command has to be.
The bot accepts any string, even roleID's, so a natural command like:
`b!editcommand setrole mute @Moderator ` would *seem* to work, but it doesn't.

It seems like a good idea to either check if the given argument is NOT a roleID in the command code, or at least mention it in the documentation and the help message.

RagingLink:
Fixes #205 
Inside `bu.hasPerm()` there is no `return Boolean` statement if the `perm` array doesn't contain a rolename, it does check and pushes roleID's to the `role` array, but it doesn't actually return anything.

This results in users not being able to provide role ID's/mentions to command requirements, as they won't return `true` inside `bu.hasPerm()`

**Added:**
- `return` statement before the boolean [31f3f41](https://github.com/blargbot/blargbot/commit/31f3f41c571a787e733955677ca9fcb8183a7f75)

### DIFF
--- a/src/utils/generic.js
+++ b/src/utils/generic.js
@@ -240,7 +240,7 @@ bu.hasPerm = async (msg, perm, quiet, override = true) => {
                 role = getId(perm);
                 if (role === null) return false;
             };
-            Array.isArray(role) ?
+            return Array.isArray(role) ?
                 role.indexOf(m.id) > -1 :
                 m.id == role;
         }


JamesZoft:
+1 to this, would be great if this was working!

PR #207, opened 2020-06-06
Rework channel related subtags

HunteRoi:
Make the BBTag API more consistent by flattening the subtags related to channels so that they act like subtags related to users and roles.

_This PR closes [feedback #551](https://blargbot.xyz/feedback/recEqELlWbyUsfv3E )._

---

**Added**
- Add `bu.getChannel` → 0e2cc87
- Add `Context.getChannel` internally using `bu.getChannel` based on the context → 0e2cc87 & 3fbed96

**Changed**
- Make `TagBuilder.util.parseChannel` asynchronous to use `context.getChannel` → 0e2cc87 & 3fbed96
- Use the asynchronous version of `Builder.util.parseChannel` in the dependent modules → 464ef56 & db734a8
- Apply DRY principle in some subtags by using `TagBuilder.withProp` → 464ef56
- Support `suppresslookup` from the context's scope → eacfd50

**Removed**
- Stop relying on `bu.parseChannel` → db734a8


HunteRoi:
@Ratismal I haven't tested anything. Also I haven't edited all JSDoc at the top of all files. You should consider removing them, they're not that much relevant, are they ?

src/structures/TagBuilder.js
```diff
@@ -257,13 +257,13 @@ TagBuilder.util = {
     error(subtag, context, message) {
         return bbEngine.addError(subtag, context, message);
     },
-    parseChannel(context, channelId) {
+    async parseChannel(context, channelId, args = { quiet: true, supppress: false }) {
         let channel = context.channel;
         if (channel.id !== channelId) {
             if (!/([0-9]{17,23})/.test(channelId))
```
RagingLink, line 263: This line checks if the channelId is either a mention or an ID, if not returns `Channel not found`, while context.getChannel does provide support for channelnames. Shouldn't it be changed to check if it's an ID, and if so set `channelId` to the matched string instead of the error?

src/tags/channelcategory.js
```diff
@@ -20,9 +20,11 @@ module.exports =
         )
         .whenArgs(0, async (_, context) => (context.channel.parentID || ''))
         .whenArgs('1-2', async (subtag, context, args) => {
-            let channel = Builder.util.parseChannel(context, args[0]);
             let quiet = bu.isBoolean(context.scope.quiet) ? context.scope.quiet : !!args[1];
-            if (typeof channel === 'function') return quiet ? '' : channel(subtag, context);
+            let channel = await Builder.util.parseChannel(context, args[0], quiet);
```
RagingLink, line 24: In this you pass a Boolean as the `args` parameter of `.parseChannel()`. An object with the quiet property should be passed instead to avoid errors to do with trying to set properties of a Boolean value in `context.getChannel()`. Could be changed to `{quiet}`.

src/tags/channeliscategory.js
```diff
@@ -20,11 +20,11 @@ module.exports =
         )
         .whenArgs(0, Builder.errors.notEnoughArguments)
         .whenArgs('1-2', async function (subtag, context, args) {
-            let channel = Builder.util.parseChannel(context, args[0]);
-
             let quiet = bu.isBoolean(context.scope.quiet) ? context.scope.quiet : !!args[1];
-            if (typeof channel === 'function') return quiet ? false : channel(subtag, context);
+            let channel = await Builder.util.parseChannel(context, args[0], quiet);
```
RagingLink, line 24: Same thing for this quiet

src/tags/channelisnsfw.js
```diff
@@ -19,15 +19,13 @@ module.exports =
             'fluffy bunnies'
         )
         .whenArgs('0-2', async function (subtag, context, args) {
-            let channel = context.channel;
-            if (args[0])
-                channel = bu.parseChannel(args[0], true);
-
             let quiet = bu.isBoolean(context.scope.quiet) ? context.scope.quiet : !!args[1];
+            let channel = context.channel;
+            if (args[0]) channel = await Builder.util.parseChannel(context, args[0], quiet);
```
RagingLink, line 24: And this quiet

src/tags/channelistext.js
```diff
@@ -19,12 +19,12 @@ module.exports =
             'nope'
         )
         .whenArgs('0-2', async function (subtag, context, args) {
-            let channel = context.channel;
-            if (args[0]) channel = Builder.util.parseChannel(context, args[0]);
-
             let quiet = bu.isBoolean(context.scope.quiet) ? context.scope.quiet : !!args[1];
-            if (typeof channel === 'function') return quiet ? false : channel(subtag, context);
+            let channel = context.channel;
+            if (args[0]) channel = await Builder.util.parseChannel(context, args[0], quiet);
```
RagingLink, line 24: This quiet too

src/tags/channelisvoice.js
```diff
@@ -19,12 +19,12 @@ module.exports =
             'nope'
         )
         .whenArgs('0-2', async function (subtag, context, args) {
-            let channel = context.channel;
-            if (args[0]) channel = Builder.util.parseChannel(context, args[0]);
-
             let quiet = bu.isBoolean(context.scope.quiet) ? context.scope.quiet : !!args[1];
-            if (typeof channel === 'function') return quiet ? false : channel(subtag, context);
+            let channel = context.channel;
+            if (args[0]) channel = await Builder.util.parseChannel(context, args[0], quiet);
```
RagingLink, line 24: Don't forget this quiet

src/tags/channelname.js
```diff
@@ -20,9 +20,10 @@ module.exports =
         )
         .whenArgs(0, async (_, context) => context.channel.name)
         .whenArgs('1-2', async (subtag, context, args) => {
-            let channel = Builder.util.parseChannel(context, args[0]);
             let quiet = bu.isBoolean(context.scope.quiet) ? context.scope.quiet : !!args[1];
-            if (typeof channel === 'function') return quiet ? '' : channel(subtag, context);
+            let channel = await Builder.util.parseChannel(context, args[0], quiet);
```
RagingLink, line 24: Another quiet D:

src/tags/channelpos.js
```diff
@@ -21,9 +21,10 @@ module.exports =
         )
         .whenArgs(0, async (_, context) => context.channel.position)
         .whenArgs('1-2', async (subtag, context, args) => {
-            let channel = Builder.util.parseChannel(context, args[0]);
             let quiet = bu.isBoolean(context.scope.quiet) ? context.scope.quiet : !!args[1];
-            if (typeof channel === 'function') return quiet ? '' : channel(subtag, context);
+            let channel = await Builder.util.parseChannel(context, args[0], quiet);
```
RagingLink, line 25: Should be the last quiet

HunteRoi:
Thanks for that. I will make the changes later! @RagingLink

RagingLink:
> Thanks for that. I will make the changes later! @RagingLink

Forgot to add this in (thanks mobile), I noticed that you aren't checking if the provided `name` is empty in `bu.getChannel(), as some subtags execute the function even if the argument is empty so you could get `undefined.toLowerCase()` for example in the filter and sort. 

Prematurely returning `null` if `name` is a falsy/not a string might be a good thing to add

src/dcommands/farewell.js
```diff
@@ -41,14 +41,15 @@ class FarewellCommand extends BaseCommand {
         let suffix = '';
         let channelStr = input.c ? input.c.join(' ') : msg.channel.id;
         if (/[0-9]{17,23}/.test(channelStr)) {
-            let channel = channelStr.match(/([0-9]{17,23})/)[1];
-            if (!bot.getChannel(channel)) {
+            const channelid = channelStr.match(/([0-9]{17,23})/)[1];
```
Ratismal, line 44: nitpick: would prefer using camelCase, so `channelId` instead of `channelid`

src/dcommands/greet.js
```diff
@@ -40,14 +40,15 @@ class GreetCommand extends BaseCommand {
         let suffix = '';
         let channelStr = input.c ? input.c.join(' ') : msg.channel.id;
         if (/[0-9]{17,23}/.test(channelStr)) {
-            let channel = channelStr.match(/([0-9]{17,23})/)[1];
-            if (!bot.getChannel(channel)) {
+            const channelid = channelStr.match(/([0-9]{17,23})/)[1];
```
Ratismal, line 43: nitpick: would prefer using camelCase, so `channelId` instead of `channelid`

src/structures/TagBuilder.js
HunteRoi: To summarize:
- `bu.getChannel` takes either a name or an id and returns the GuildChannel object
- `bu.parseChannel` takes a text and checks for an id (resolved from a mention within the text) and returns the GuildChannel object
- `Builder.util.parseChannel` takes a name or an id (or a mention before changing it to an id). If it can resolve the channel, it returns the GuildChannel object. If it can't, it feeds `context.getChannel` with it.
- `context.getChannel` takes a name or an id. If it can resolve the GuildChannel object, it does. If it can't, it feeds `bu.getChannel` with the data

HunteRoi: _Check my last commit for the changes :wink:_

HunteRoi:
ew I rebased my branch on master, what a mistake !

Ratismal:
Oh no im sorry >.<

HunteRoi:
@Ratismal it's ready for regression tests and touchups

HunteRoi:
:up:

RagingLink:
@HunteRoi I looked into this PR just now, as I thought it could possibly fix `channel.permissionsOf is not a function`, and I noticed that one half of the subtags use `if (!channel)` and the other half use `if (typeof channel === 'function')` (not sure which one is supposed to be used tbh). `TagBuilder.util.parseChannel()` seems to suggest it's the latter. Could you update this PR to correct the subtags that still use the old method?

**`if (!channel)`**:
- `channeldelete, channeledit, channelsetperms, channelsetpos, reactadd, reactlist, reactremove, reactremoveall, send and slowmode`

**`if (typeof channel === 'function')`:**
- `channelcategory, channeliscategory, channelisnsfw, channelisttext, channelisvoice, channelname, channelpos, channels, channeltype, delete`

HunteRoi:
> @HunteRoi I looked into this PR just now, as I thought it could possibly fix `channel.permissionsOf is not a function`, and I noticed that one half of the subtags use `if (!channel)` and the other half use `if (typeof channel === 'function')` (not sure which one is supposed to be used tbh). `TagBuilder.util.parseChannel()` seems to suggest it's the latter. Could you update this PR to correct the subtags that still use the old method?
> 
> **`if (!channel)`**:
> 
> * `channeldelete, channeledit, channelsetperms, channelsetpos, reactadd, reactlist, reactremove, reactremoveall, send, edit and slowmode`
> 
> **`if (typeof channel === 'function')`:**
> 
> * `channelcategory, channeliscategory, channelisnsfw, channelisttext, channelisvoice, channelname, channelpos, channels, channeltype, delete`

I'm actually not sure the latter is correct. A channel instance is not supposed to be of type function but rather of object. Checking if channel is not defined can be done in two ways:
- `if (!channel)`
- `if (typeof channel === 'undefined')`
The issue with the typeof check is that you don't take into account that the variable can actually hold a nullable value which then results in a type of "object".

I think the best way of checking if a channel is properly returned in those tags are to verify that we have none of the values that can make the code error out, so to speak: null, undefined or anything that is not a Channel.
As we call a custom method, we could control that we either receive a null or Channel instance, which would then simplify the condition to `if (!channel)` only.

IMO the tags that need to be edited are the ones using the function typeof comparison.

RagingLink:
I agree that `if (!channel)` is cleaner than the whole function check and thus the function return of `.parseChannel`. Thank you for specifying what should be used in the channel subtags, I'll look into making the necessary changes so this can (finally) be merged.

RagingLink (approved):

RagingLink (approved):

PR #211, opened 2020-06-30
Fixed arrays being stored as string with jset

RagingLink:
Resolves [Airtable report](https://airtable.com/shrEUdEv4NM04Wi7O/tblyFuWE6fEAbaOfo/viwDg5WovcwMA9NIL/recDo2XJxNngfJCHi)

Todo:
- [ ] Properly handle input array as `obj`

**Added**:
- deserialization of array values [c2c8fb8](https://github.com/blargbot/blargbot/commit/c2c8fb89daa933cd8b6b5f11e81b052977b71fea)

### DIFF
--- a/src/tags/jsonset.js
+++ b/src/tags/jsonset.js
@@ -1,8 +1,8 @@
 /*
  * @Author: stupid cat
  * @Date: 2017-05-07 18:49:14
- * @Last Modified by: stupid cat
- * @Last Modified time: 2018-09-06 14:19:25
+ * @Last Modified by: RagingLink
+ * @Last Modified time: 2020-06-30 18:45:41
  *
  * This project uses the AGPLv3 license. Please read the license file before using/adapting any of the code.
  */
@@ -29,7 +29,10 @@ module.exports =
                 obj = '{}';
 
             let varname = undefined;
-
+            const deserializedTagArray = await bu.deserializeTagArray(value);
+            if (deserializedTagArray)
+                value = deserializedTagArray.v;
+            
             let arr = await bu.getArray(obj);
             if (arr && Array.isArray(arr.v)) {
                 obj = arr.v;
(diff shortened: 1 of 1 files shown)


src/tags/jsonset.js
```diff
@@ -30,7 +30,7 @@ module.exports =
 
             let varname = undefined;
             const deserializedTagArray = await bu.deserializeTagArray(value);
-            if (deserializedTagArray)
+            if (deserializedTagArray && !deserializedTagArray.n)
```
HunteRoi, line 33: Should be `if (deserializedTagArray && deserializedTagArray.n)` as you are checking for a value in the variable `deserializedTagArray` but also checking that this object has a property `n` different from a falsy value.

RagingLink: I actually thought the same, but if the input array is of the format {"v": ["1","2","3"], "n": "~array"}, it will return the same input object but just parsed. 

You want to make sure that the input array isn't a 'variable array (to keep the reference to the variable, and allow modifications to the array similar to how JS works with references)', and that's why I added `!deserializedTagArray.n`, as that will return a falsy if it doesn't have a name (and thus isn't an array variable).

Like I mentioned on Discord, if you want to pass on an array variable without keeping the reference to the variable, you'd use `{slice;{get;~array};0}.

HunteRoi: So what has been decided?

RagingLink: Keeping the if statement as is? I'm not entirely sure what's the remaining issue. Your request says it should be acceptable for arrays to be passed without name (not a variable), and that's kind off the only thing that this if statement handles. It only accepts arrays passed without name.

RagingLink:
I don't think this one quite works out, objects should be also be included as these are currently stored as string too. 
I'll have to change this to work for both objects and arrays. 
```js
const deserializedTagArray = await bu.deserializeTagArray(value);
if (deserializedTagArray) { 
  value = deserializedTagArray.v;
} else {
  try {
    value = JSON.parse(value);
  } catch(e) { };
};
```
I'm thinking of changing it to the above code, but I'll have to do some testing first

PR #212, opened 2020-06-30
New escapebbtag subtag

RagingLink:
New subtag `escapebbtag` with alias `escape`. This would provide an easy way of escaping BBtag code, without needing to use `{rb}`, `{lb}` and `{semi}`. This could be useful if you want to set a variable to something that you don't want to parse as BBtag when setting it (without using character escape subtags).
 Based on the following suggestions: [582](https://airtable.com/shrEUdEv4NM04Wi7O/tblyFuWE6fEAbaOfo/viwl0N62QIReJLMJv/reckvBdm7skJUgIUp), [549](https://airtable.com/shrEUdEv4NM04Wi7O/tblyFuWE6fEAbaOfo/viwl0N62QIReJLMJv/recGmj25ttQWk3VE3), [532](https://airtable.com/shrEUdEv4NM04Wi7O/tblyFuWE6fEAbaOfo/viwl0N62QIReJLMJv/recoukHUHMAm2dKa5), [160](https://airtable.com/shrEUdEv4NM04Wi7O/tblyFuWE6fEAbaOfo/viwl0N62QIReJLMJv/recjGBMJP02gWH0vf)

**Added**:
- escapebbtag subtag [459d18c](https://github.com/blargbot/blargbot/commit/459d18c415cb7987ac613b8534a65d1aed2c7572)

### DIFF
--- a/src/tags/escapebbtag.js
+++ b/src/tags/escapebbtag.js
@@ -2,7 +2,7 @@
  * @Author: RagingLink 
  * @Date: 2020-06-25 12:25:05
  * @Last Modified by: RagingLink
- * @Last Modified time: 2020-06-30 21:59:10
+ * @Last Modified time: 2021-05-21 00:05:34
  *
  * This project uses the AGPLv3 license. Please read the license file before using/adapting any of the code.
  */
@@ -14,7 +14,8 @@ module.exports =
         .withArgs(a => [a.require('code')])
         .withAlias('escape')
         .withDesc('Returns `code` without executing any BBtag inside it.\n' +
-            'This effectively returns the characters `{`,`}` and `;` as is, without the use of `{rb}`, `{lb}` and `{semi}`.')
+            'This effectively returns the characters `{`,`}` and `;` as is, without the use of `{rb}`, `{lb}` and `{semi}`.\n' +
+            '**NOTE**: Brackets inside `code` must come in pairs. A { *has to be* followed by a } somewhere and a } *has to have* a { before it')
         .withExample(
             '{escapebbtag;{set;~index;1}}',
             '{set;~index;1}'
--- a/src/tags/escapebbtag.js
+++ b/src/tags/escapebbtag.js
@@ -1,0 +1,27 @@
+/**
+ * @Author: RagingLink 
+ * @Date: 2020-06-25 12:25:05
+ * @Last Modified by: RagingLink
+ * @Last Modified time: 2020-06-30 21:59:10
+ *
+ * This project uses the AGPLv3 license. Please read the license file before using/adapting any of the code.
+ */
+
+const Builder = require('../structures/TagBuilder');
+
+module.exports =
+    Builder.BotTag('escapebbtag')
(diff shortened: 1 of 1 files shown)


src/tags/escapebbtag.js
```diff
@@ -0,0 +1,27 @@
...
+        .withDesc('Returns `code` without executing any BBtag inside it.\n' +
+            'This effectively returns the characters `{`,`}` and `;` as is, without the use of `{rb}`, `{lb}` and `{semi}`.')
+        .withExample(
+            '{escapebbtag;{set;~index;1}}',
+            '{set;~index;1}'
+        )
+        .resolveArgs(-1)
+        .whenArgs(0, Builder.errors.notEnoughArguments)
+        .whenDefault(async (subtag, context, args) => {
+            return args.map(arg => arg.content).join(';');
```
HunteRoi, line 25: It looks too easy to me, but I cannot wrap my head around it. Will take another look later.

RagingLink: If you want I could explain it a bit 🤷‍♂️

HunteRoi: I'd be happy if you could tell me more

RagingLink: `.resolveArgs(-1)` makes it so blargbot doesn't parse the BBtag inside the subtag, and thus `args` is an array of BBtag instances. Every BBtag instance has a `.content` property with the 'raw string' of everything inside it. The map just changes the args array so every element is the `content` of the argument. 
However since blargbot counts `;` as an argument seperator (`{escape;Arg1;Arg2;Arg3}`) you will have `["Arg1","Arg2","Arg3"]` as the result of `args.map(...)`, by doing `.join(';')` you effectively get those semicolons back where they once were. Any whitespace characters inside the subtag will just be the same.

HunteRoi: But what happens if for instance I input this :
```csharp
{escapebbtag;
  {send;{get;~channelId};
    This is a long message

with a few newlines
                and a few whitespaces
    just because I can do it
  }
}
```

RagingLink: Whitespace characters only get 'removed' when parsing the first  *official* argument of a subtag (the name), and the following arguments **if** they're executed. Since `.resolveArgs(-1)` is specified they're not. So any whitespace character inside `code` remains there, and that includes other "subtags" inside it. Even the extra newlines you have in the beginning and end will not get removed.

Ratismal: Something to note here is that this will only work if the contents contain matched sets of { and }. If the quantities are uneven, or if } appears before {, it may behave unexpectedly. That's the main reason why this behaviour hasn't been implemented to begin with.

This subtag is probably worth having, but this should be mentioned as a caveat.

RagingLink: I guess that should be documented. I remember thinking about this exact issue and how it could be "solved" but I couldn't come up with something without having to change the way subtags are parsed.

HunteRoi (approved):

PR #213, opened 2020-07-01
Added new /json endpoint for shards

RagingLink:
Similar to how `/domain` has  `/json`, `/shards` now has `/json`. This would make viewing shards/cluster info a lot easier via HTTP(s) requests.

**Added**:
- New `json` endpoint for shards [9568a8f](https://github.com/blargbot/blargbot/commit/9568a8f418bd306fe47cad3a15e41773ae376272)
- shards.js router file

**Changed**:
- Shards endpoint is now in a seperate router instead of `index.js`


src/backend/routes/shards.js
```diff
@@ -0,0 +1,23 @@
...
+
+router.get('/', (req, res) => {
+    res.locals.url = config.general.isbeta ? 'ws://localhost:8085' : 'wss://blargbot.xyz';
+
+    res.render('shards');
+});
+
+router.get('/json', (req, res) => {
+    res.type('json');
+    res.send(JSON.stringify(Object.values(spawner.shardCache), null, 2));
```
HunteRoi, line 20: Are you willing to send only shards information as an array of arrays ? Wouldn't it be better to have shards per clusters?

RagingLink: https://blargbot.xyz/output/727968583769981940/raw
That's the output of `Object.values(spawner.shardCache)`, it's an array with all cluster objects, and a cluster object has a shards key with an array of shard objects

HunteRoi: IMAGE

That's quite ugly imo

RagingLink: The shardCache should really be called `clusterCache`, as that's what it is.

HunteRoi: You could make that change too by the way :P

Preferably, I'd rather receive something like this if requesting `/shards/json`
IMAGE

RagingLink: I mean, does the `shardCache` need to prettified? With `/json` you just want to get the `raw` cluster information, right?

RagingLink: > You could make that change too by the way :P
> 
> Preferably, I'd rather receive something like this if requesting `/shards/json`
> IMAGE

For that instead of `Object.values()` it'd just need to return `.shardCache` directly. Unless you're referring to the actual display of it?

HunteRoi: I think it's better to return shardCache instead of Object.values result

PR #216, opened 2020-07-17
New map subtag for arrays

RagingLink:
Created a map subtag that maps the input array and either returns the output array, or modifies the original array if a variable is provided. 

I'm still not sure if the limits are appropriate, currently `map` has the same limits as `foreach`, as code-wise they're very similar.

**Added**:
- Map subtag [693c1b0](https://github.com/blargbot/blargbot/commit/693c1b021751b28883712bbe297d1e76bf53ee54)
- Map limits [57c44fa](https://github.com/blargbot/blargbot/commit/57c44fab88072dc473ef2b91bb9392ecf1429d6d)

### DIFF
--- a/src/tags/map.js
+++ b/src/tags/map.js
@@ -23,7 +23,7 @@ module.exports =
             ' it will simply output the new array.'
         )
         .withExample(
-            '{set;~array;apples;oranges;pears}\n{map;~item;{get;~array};{upper;{get;~item}}',
+            '{map;~item;["apples","oranges","pears"];{upper;{get;~item}}}',
             '["APPLES","ORANGES","PEARS"]'
         )
         .resolveArgs(0, 1)
--- a/src/tags/map.js
+++ b/src/tags/map.js
@@ -2,7 +2,7 @@
  * @Author: RagingLink 
  * @Date: 2020-06-03 23:26:05
  * @Last Modified by: RagingLink
- * @Last Modified time: 2020-06-04 00:17:35
+ * @Last Modified time: 2020-07-17 22:17:35
  *
  * This project uses the AGPLv3 license. Please read the license file before using/adapting any of the code.
  */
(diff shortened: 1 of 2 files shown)


src/tags/map.js
```diff
@@ -0,0 +1,58 @@
...
+            '{map;~item;["apples","oranges","pears"];{upper;{get;~item}}}',
+            '["APPLES","ORANGES","PEARS"]'
+        )
+        .resolveArgs(0, 1)
+        .whenArgs('0-2', Builder.errors.notEnoughArguments)
+        .whenArgs('3', async function (subtag, context, args) {
+            let varName = args[0],
+                arr = await bu.getArray(context, args[1]),
+                array = Array.from(arr.v),
+                newArray = [];
```
HunteRoi, line 35: I'm not a fan of declaration with `,` but it's ok I guess

src/tags/map.js
```diff
@@ -0,0 +1,58 @@
...
+            
+            let remaining = context.state.limits.map || {loops : NaN};
```
HunteRoi, line 37: Why would you set remaining with `{ loops: NaN }` ?

src/tags/map.js
```diff
@@ -0,0 +1,58 @@
...
+            for (const item of array) {
+                remaining.loops--;
+                if (!(remaining.loops >= 0)) {
```
HunteRoi, line 40: reverse the condition : `remaining.loops < 0`

src/tags/map.js
RagingLink: Think I agree on that, though since most subtags also use this I figured it'd be ok

src/tags/map.js
RagingLink: Honestly not so sure about that one. I try to maintain/reflect the code of other subtags as much as possible and other loop subtags use `{loops : NaN}` too 🤔
Should be able to remove it though, as `.limits.map` *should* always be defined, but then I'm not sure why other subtags **do** have it

src/tags/map.js
RagingLink: ```// (remaining.loops < 0) would not work due to the comparison behaviours of NaN``` - Comment from `{repeat}`
If I'd remove the `{loops : NaN}`, changing it to `remaining.loops < 0` would be a lot cleaner. But as I mentioned in my previous reply, I'm not even sure if I should remove it.

Ratismal:
Note: I'm going to adjust this so that it doesn't mutate the original array, to be more consistent with the js map function thats linked in the description.

PR #228, opened 2021-01-03
Fixed issue with filter and arrays with objects

RagingLink:
Fixes bug report [595](https://airtable.com/shrEUdEv4NM04Wi7O/tblyFuWE6fEAbaOfo/viwDg5WovcwMA9NIL/recTKg4fJgYCobLIb?blocks=bipip3uBZFRPcSy3C)

To avoid storing objects in the `processed` object as `[object Object]`, they need to be stringified when checking if it's processed, when storing them and when pushing equal objects to `result`.

### DIFF
--- a/src/tags/filter.js
+++ b/src/tags/filter.js
@@ -39,7 +39,8 @@ module.exports =
             let i = 0;
 
             for (const item of array) {
-                if (processed[item]) continue;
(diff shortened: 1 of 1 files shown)


Ratismal:
Hmm... my first instinct is that I'd prefer to get a hash from the object and use that as an index, but also that might be stupid and this is probably fine.

PR #233, opened 2021-01-28
Sort prefixes with longer prefixes first

RagingLink:
Sorts prefixes based on the length instead of `a.length < b.length`, this is done in descending order so in cases where a user has a blank prefix `""` it would always pick that one and render other prefixes practically useless (as blargbot tries to execute a command with the blank prefix)
In response to suggestion [689](https://airtable.com/shrEUdEv4NM04Wi7O/tblyFuWE6fEAbaOfo/viwWwUA753iEQqOo7/recsAvekO5oAM2YVQ?blocks=bipip3uBZFRPcSy3C)

**Changed:**
- Sorting function from `a.length < b.length` to `b.length - a.length` [a247b95](https://github.com/blargbot/blargbot/commit/a247b953db088f6a26b903163032912f5208d99c)

### DIFF
--- a/src/events/messageCreate.js
+++ b/src/events/messageCreate.js
@@ -60,7 +60,7 @@ async function handleUserMessage(msg, storedGuild) {
     };
     prefixes.push(config.discord.defaultPrefix, 'blargbot');
     prefixes.sort((a, b) => {
-        return a.length < b.length;
+        return b.length - a.length; //Sort descending
     });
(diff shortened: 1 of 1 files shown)


HunteRoi (approved):

HunteRoi:
@Ratismal you can close the ticket 689 when merging this PR

Ratismal:
I could have sworn that I had fixed this... but I guess it's been in this state for the past 3 years? Super confused :/

PR #235, opened 2021-01-31
Add zws character so blarg strips the beginning mention in cleverbot

RagingLink:
As of Eris 14.0.0 they replace mentions with ` @' + '\u200b' + username` (changed in commit [13390b9](https://github.com/abalabahaha/eris/commit/13390b9b5328f57173b4a6ae4b2487ead9f1f386 "Escape more cleanContent mentions")). Since blargbot isn't checking for the zws character in the regex, the mention doesn't get stripped and results in random responses instead of the usual "Hi. My name is blargbot..." `
Examples: 
image

**Added**:
- `\u200b` to the Regex `Regexp('@' + username + ',?')/` [e8df93e](https://github.com/blargbot/blargbot/commit/e8df93ecc46cbe9ec31735abb68a7c1e9c86db85 "Add zws to stripped mention cleverbot")

### DIFF
--- a/src/events/messageCreate.js
+++ b/src/events/messageCreate.js
@@ -583,7 +583,7 @@ async function handleCleverbot(msg) {
     var username = msg.channel.guild.members.get(bot.user.id).nick ?
         msg.channel.guild.members.get(bot.user.id).nick :
         bot.user.username;
-    var msgToSend = msg.cleanContent.replace(new RegExp('@' + username + ',?'), '').trim();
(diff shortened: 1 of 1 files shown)


Ratismal:
Hmm, so this one confuses me. I don't have have a prewritten message for `@blargbot Hello.`, so that's purely on the API that I'm using.

What I'm more confused about is the fact that it seems to be responding to every single sentence individually? has that always been the case?

I need to find a better bot >.> maybe host my own gpt-2... hahaaaaaaa.......

PR #237, opened 2021-02-02
Move all magic discord strings to config

danny-may:


HunteRoi (commented):
Seems good. @Ratismal there are a few `// TODO` marking unknown fields. Would be interesting to take a look at these :)

By the way, hope you guys are doing great! Have a good evening :smile:

danny-may:
Already had a chat with cat about those for the typescript PR i have in draft. Im not sure this one will make sense to be merged any more tbh

PR #248, opened 2021-05-11
Added support for flag names in exec and execcc

RagingLink:
Changed `exec` and `execcc` to allow usage of flag names if they've been added. 
**Example:**
`tag1` with `b!t flag add tag1 -f flag`:
```csharp
{flagset;f}
```
`tag2`
```csharp
{exec;tag1;{args}}
```
`b!t tag2 --flag` returns `true`
Current behaviour is seen in the screenshot below
image

I also added a comment to a piece of code that can be removed, as mentioned in the comment itself.
**Added:**
- `flags` parameter to the `execTag` method in `{exec}`
- `.flags` value as argument in `{exec}` and `{execcc}` [8f6f179](https://github.com/blargbot/blargbot/commit/8f6f17981dbc168c6283f0122a0280316a0bc4a5)

### DIFF
--- a/src/tags/execcc.js
+++ b/src/tags/execcc.js
@@ -1,8 +1,8 @@
 /*
  * @Author: stupid cat
  * @Date: 2017-05-07 18:37:21
- * @Last Modified by: stupid cat
- * @Last Modified time: 2018-06-04 11:39:51
+ * @Last Modified by: RagingLink
+ * @Last Modified time: 2021-05-11 18:54:06
  *
  * This project uses the AGPLv3 license. Please read the license file before using/adapting any of the code.
  */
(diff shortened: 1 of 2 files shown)


src/tags/execcc.js
```diff
@@ -46,11 +46,12 @@ module.exports =
                 case 1:
                     return TagManager.list['exec'].execTag(subtag, context, ccommand.content, '');
                 case 2:
-                    return TagManager.list['exec'].execTag(subtag, context, ccommand.content, args[1]);
+                    return TagManager.list['exec'].execTag(subtag, context, ccommand.content, args[1], ccommand.flags);
```
HunteRoi, line 49: Are the flags provided when calling the ccommand possibly contented in the length of `args` ?

HunteRoi: Is it possible that `args.length` also takes into account the flags? Because it would then change the number of detected arguments then the `case 2:` changes may be useless.

RagingLink: Ah, I think I understand your question ow, but I don't quite understand the confusion tbh 🤔 
`args.length` includes the tag/ccommand name and every argument after it. So `b!foo --bar` would have an `args.length` of `2` an be `case 2:` (and thus require `ccommand.flags` to be passed to recognise the `--bar` flag).

HunteRoi (approved):

PR #249, opened 2021-05-20
Added bitwise subtag

RagingLink:
I made a while ago but now got remembered to create a PR so this can get added to blarg.

This is a bitwise subtag that follows the same logic of `{math}` and supports the following operators: `&, |, ^, ~, <<, >>, >>>` .

If necessary the error message when providing an invalid binary number can be changed to `Not a binary number` with help of `.customError(...)`.

### DIFF
--- a/src/tags/bitwise.js
+++ b/src/tags/bitwise.js
@@ -1,0 +1,48 @@
+/**
+ * @Author: RagingLink 
+ * @Date: 2020-08-17 14:03:17
+ * @Last Modified by: RagingLink
+ * @Last Modified time: 2021-05-20 23:03:10
+ *
+ * This project uses the AGPLv3 license. Please read the license file before using/adapting any of the code.
+ */
+
+ const Builder = require('../structures/TagBuilder');
+ let operators = {
+     '&': (a, b) => a & b,
+     '|': (a, b) => a | b,
+     '^': (a, b) => a ^ b,
+     '~': (a) => ~a,
+     '<<': (a, b) => a << b,
+     '>>': (a, b) => a >> b,
+     '>>>': (a, b) => a >>> b
+ }
+ 
+ module.exports =
+     Builder.AutoTag('bitwise')
+         .acceptsArrays()
+         .withArgs(a => [a.require('operator'), a.require('value'), a.optional('othervalue', true)])
+         .withAlias('bitmath')
+         .withDesc(`Performs a bitwise operation on the provided value(s). Supplied values must be binary numbers.\nValid operators are: \`${Object.keys(operators).join(', ')}\``)
+         .withExample(
+             '{bitwise;^;1000;1001}',
+             '1000'
+         ).whenArgs('0-1', Builder.errors.notEnoughArguments)
+         .whenDefault(async (subtag, context, args) => {
+             if (!operators.hasOwnProperty(args[0]))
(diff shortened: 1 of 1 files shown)


src/tags/bitwise.js
```diff
@@ -0,0 +1,48 @@
...
+ * @Author: RagingLink 
+ * @Date: 2020-08-17 14:03:17
+ * @Last Modified by: RagingLink
+ * @Last Modified time: 2021-05-20 23:03:10
+ *
+ * This project uses the AGPLv3 license. Please read the license file before using/adapting any of the code.
+ */
+
+ const Builder = require('../structures/TagBuilder');
+ let operators = {
```
HunteRoi, line 11: 2 things here ; the variable can be declared as a constant **and** the naming bugs me. I know it's because it is not only operators in the object but rather like an `operatorsToOperations` dictionary. Do you think we can find a better name?

src/tags/bitwise.js
```diff
@@ -0,0 +1,48 @@
...
+         .withAlias('bitmath')
+         .withDesc(`Performs a bitwise operation on the provided value(s). Supplied values must be binary numbers.\nValid operators are: \`${Object.keys(operators).join(', ')}\``)
+         .withExample(
+             '{bitwise;^;1000;1001}',
+             '1000'
+         ).whenArgs('0-1', Builder.errors.notEnoughArguments)
+         .whenDefault(async (subtag, context, args) => {
+             if (!operators.hasOwnProperty(args[0]))
+                 return Builder.errors.invalidOperator(subtag, context);
+             let operator = operators[args[0]]
```
HunteRoi, line 34: I'd rather call the variable `operation` as the object returns an operation based on the provided operator.

src/tags/bitwise.js
```diff
@@ -0,0 +1,48 @@
...
+             
+             if (args[0] === '~')
+                 return operator(parseInt(args[1], 2)).toString(2);
+             if(args.length === 2)
```
HunteRoi, line 38: You're just missing a space here (consistency is key :smile:)

src/tags/bitwise.js
```diff
@@ -0,0 +1,48 @@
...
+             '{bitwise;^;1000;1001}',
+             '1000'
+         ).whenArgs('0-1', Builder.errors.notEnoughArguments)
+         .whenDefault(async (subtag, context, args) => {
+             if (!operators.hasOwnProperty(args[0]))
+                 return Builder.errors.invalidOperator(subtag, context);
+             let operator = operators[args[0]]
+             
+             if (args[0] === '~')
+                 return operator(parseInt(args[1], 2)).toString(2);
```
HunteRoi, line 37: This is a pure subjective comment but for readability. I would externalize the `parseInt(args[1], 2)` to a function `convertToBinary` or `parseToBase2` or something like that.
There's also a method `Number.prototype.toPrecision` that does the same thing as `toString(2)` without changing "its type" : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toPrecision

Changes like that would give us:
```js
if (args[0] === '~') {
  return operator(convertToBinary(args[1])).toPrecision(2);
}
```

HunteRoi (changes requested):

PR #250, opened 2021-05-20
Possible fix for internal errors coming from rolecreate and rolesetperms

RagingLink:
`{rolecreate}` and `{rolesetperms}` seem to be returning `Internal server occurred`  every single time and from my testings it seems to be because of the error `Cannot mix BigInt and other types, use explicit conversions` on https://github.com/blargbot/blargbot/blob/4d9d40a65efbf5986c073abd101d410bdb0645ff/src/tags/rolecreate.js#L50 and https://github.com/blargbot/blargbot/blob/4d9d40a65efbf5986c073abd101d410bdb0645ff/src/tags/rolesetperms.js#L38

Because these are internal errors I can only try to guess what's going wrong on blarg, but I'm fairly certain that this is the reason of the internal server error

**Changed**:
- Convert `options.permissions` to BigInt [4a2b8a7](https://github.com/blargbot/blargbot/commit/4a2b8a76dcbe7d6d4704161cf759ed43da8ae5ac)
- Convert `perms` to BigInt [4f60c85](https://github.com/blargbot/blargbot/commit/4f60c85788e623feb0423e4078673a016654ea69)

### DIFF
--- a/src/tags/rolecreate.js
+++ b/src/tags/rolecreate.js
@@ -1,8 +1,8 @@
 /*
  * @Author: stupid cat
  * @Date: 2017-05-21 00:22:32
- * @Last Modified by: stupid cat
- * @Last Modified time: 2019-09-26 09:29:09
+ * @Last Modified by: RagingLink
+ * @Last Modified time: 2021-05-20 23:45:03
  *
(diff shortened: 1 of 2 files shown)


HunteRoi (approved):
LGTM!

PR #251, opened 2021-06-10
Typescript conversion

RagingLink:
**Added:**
- Required property `usage` in `BaseSubtag`.
- eslint rule to allow `//@ts-...` comments.
- `customError` and `notEnoughArguments` errors in `BaseSubtag`.
- TypeScript versions of a couple subtags.

**Changed:**
- Behaviour when using `whenArgCount : { '1': (...) => ...}`, instead of changing the argument range to include all possible lengths, this only changes `minCount` and `maxCount` to `1` (or whatever argCount for that matter).


HunteRoi:
Sorry to bump here, but there are a few missing things here and there, at least the two missing brackets I've seen in base64decode & base64encode's exampleCode value

HunteRoi:
Couldn't review much of the PR before @danny-may merged it so I guess it's still worth mentioning it for later ;) have a nice day guys :D

RagingLink:
> Sorry to bump here, but there are a few missing things here and there, at least the two missing brackets I've seen in base64decode & base64encode's exampleCode value

I'll take a look at that later! I might have missed a few when checking them. Thank you for noticing and pointing it out o/

PR #256, opened 2021-06-11
Couple subtags converted to TS

RagingLink:
**Added:**
- `{bool}, {lock}, {set}, {get}, {json}`

**Changed:**
- When getting variables automatically try to `JSON.parse`, when storing them try `JSON.stringify` (old behaviour no biggie)
- Location of various utility functions

TODO:
- [ ] Properly implement `ReadWriteLock` with typed indexes so it's seen as 'valid' by TS, couldn't quite come up with something.


RagingLink:
Spotted a few log statements for debugs, oops!

src/tags/lock.ts
```diff
@@ -0,0 +1,108 @@
...
+        if (scope == null) throw new Error('Missing default variable scope!');
+
+        let lock = scope.getLock(
+            context,
+            subtag,
+            key.substring(scope.prefix.length)
+        );
+        //TODO fix this type error
+        //@ts-ignore
+        let lockFunc = lock[mode + 'Lock'].bind(this);
```
RagingLink, line 84: ❌ .bind

src/tags/bool.ts
```diff
@@ -0,0 +1,129 @@
...
+                            args[1].value,
+                            args[2].value
+                        )
+                }
+            ]
+        });
+        //TODO possibly remove this?
+        this.operators = operators;
+        /**
+         * * There has to be a way to better define these types??
```
danny-may, line 53: I agree, these dont need to be here.  One idea cat and I were throwing around before was an operator subtag
`{>;x;y}` something like this which can be referenced by `{bool}`, `{logic}`, `{if}`, `{math}` etc. Maybe this would be a good time to implement this?
Move the operator logic to util, have the operators grouped by category (bool, numeric, compare).

`{bool}` and `{if}` would use the `compare` operators (`==`, `!=`, `<`, `>` etc.)
`{logic}` would use the `bool` operators (`||`, `&&`, `!` etc)
`{math}` would use the `numeric` operators (`+`, `-`, `*` etc)
`{operator}` would use all the operators (`{+;5;6;7;8}`, `{*;10;9;8;7}`, `{!;true}`, `{>;a;b}` etc

src/tags/bool.ts
```diff
@@ -0,0 +1,129 @@
...
+                    description:
+                        'Evaluates `arg1` and `arg2` using the `evaluator` and returns `true` or `false`. ' +
+                        'Valid evaluators are `' +
+                        operators.join('`, `') +
+                        '`\n' +
+                        'The positions of `evaluator` and `arg1` can be swapped.',
+                    exampleCode: '{bool;5;<=;10}',
+                    exampleOut: 'true',
+                    execute: (ctx, args, subtag) =>
+                        this.runCondition(
```
danny-may, line 40: Can we try to keep these on 1 line. I think its fine in this situation if it ends up being a long line as there should be little to no logic in these, its just a redirection to an actual method below so keeping the definition section easy to scan over is more important I think

src/tags/if.ts
```diff
@@ -0,0 +1,157 @@
...
+                    description:
+                        'If `boolean` is `true`, return `then`, else do nothing.',
+                    execute: (ctx, [{ value: bool }, thenCode], subtag) =>
+                        this.simpleBooleanCheck(ctx, subtag, bool, thenCode)
+                },
+                {
+                    args: ['boolean', '~then', '~else'],
+                    description:
+                        'If `boolean` is `true`, return `then`, else execute `else`',
+                    execute: (
```
danny-may, line 42: Same as above, can we keep these to 1 line

src/tags/set.ts
```diff
@@ -0,0 +1,95 @@
...
+                        await ctx.variables.set(variableName, undefined)
+                },
+                {
+                    args: ['name', 'value'],
+                    description:
+                        'Stores `value` under `name`. These variables are saved between sessions. ' +
+                        'You can use a character prefix to determine the scope of your variable.\n' +
+                        'Valid scopes are: ' +
+                        tagVariableScopes
+                            .map(
```
danny-may, line 28: Same as above, keep this to 1 line if possible

src/tags/bool.ts
RagingLink: Sounds like a really good idea, I will be sure to update this PR with it when I have the time.

For future commands/subtags, should 'shared' code just always be moved to utils so you wouldn't have stuff like `this.subtags.get('bool').methodHere()` or like `this.commands.get('ban').ban()`?

src/tags/bool.ts
RagingLink: Sure!

src/tags/bool.ts
danny-may: I think if its shared code and doesnt rely on anything else (discord, database etc) then yes, move to util following the pattern there. otherwise, if its only used on the clusters move it to ClusterUtilities, if its used by the master process and the clusters then move it to BaseUtilities

src/tags/set.ts
```diff
@@ -0,0 +1,95 @@
...
+                    exampleCode:
+                        '{set;var1;This is local var1}\n{set;~var2;This is temporary var2}\n{get;var1}\n{get;~var2}',
+                    exampleOut: 'This is local var1\nThis is temporary var2',
+                    execute: async (
+                        ctx,
+                        [{ value: variableName }, { value }]
+                    ) => await this.set(ctx, variableName, value)
+                },
+                {
+                    args: ['name', 'element1', 'otherElements+'],
```
danny-may, line 48: `[`name', 'values+2']` 😉

RagingLink: I'm an idiot, thank you

danny-may (approved):

PR #262, opened 2021-06-19
Added suppresslookup for channel subtags

RagingLink:
Allow users to suppress the "No channels found" output from the lookup system, because of [this discussion](https://discord.com/channels/194232473931087872/238131984017260546/855639759347122186)

**Added:**
- Pass on `context.scope.suppressLookup` to channel lookups

### DIFF
--- a/src/tags/channelcategory.js
+++ b/src/tags/channelcategory.js
@@ -2,7 +2,7 @@
  * @Author: zoomah
  * @Date: 2018-07-10 7:08:15
  * @Last Modified by: RagingLink
- * @Last Modified time: 2021-06-13 14:59:41
+ * @Last Modified time: 2021-06-19 10:25:36
  *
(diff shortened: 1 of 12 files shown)


danny-may (approved):

HunteRoi:
Just a side note from someone who's following the changes away from the Discord guild: maybe it would be good to summarize any topic when it is subject to a PR and link a Discord's discussion as "source" or "more info at" for details. What do you think? @RagingLink

RagingLink:
> Just a side note from someone who's following the changes away from the Discord guild: maybe it would be good to summarize the topic when it's subject to a PR and link a Discord's discussion as "source" or "more info at" for details. What do you think? @RagingLink 

Didn't think about this to be honest, but yeah I agree.  I'll keep this in mind for the next time! Thank you👍🏻

PR #273, opened 2021-08-02
Added more transparent information regarding timeouts

RagingLink:
Seeing as the command timeout isn't really documented anywhere, I added some more info to the 'timeout message'.

**NOTE**: I also noticed that the `penalty` variable is unused (kept it that way), but it might be good to keep this in mind in case the penalty gets modified in the future.

### DIFF
--- a/src/events/messageCreate.js
+++ b/src/events/messageCreate.js
@@ -2,7 +2,7 @@
  * @Author: stupid cat
  * @Date: 2017-05-07 18:22:24
  * @Last Modified by: RagingLink
- * @Last Modified time: 2021-08-02 18:34:43
+ * @Last Modified time: 2021-08-03 12:08:32
  *
  * This project uses the AGPLv3 license. Please read the license file before using/adapting any of the code.
  */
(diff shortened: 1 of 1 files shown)


danny-may:
On the fence about this one. If we tell people why they have been timed out then they might make a workaround to ride the limit. On the other hand, blarg is open source so they could just have a look at the code to figure it out...

15commands/30s is a pretty generous limit, if someone hits that and they cant figure out that they were spamming then i dont know what to say lol

RagingLink:
> On the fence about this one. If we tell people what the limit is then they might make a workaround to ride the limit. On the other hand, blarg is open source so they could just have a look at the code to figure it out...
> 
> 15commands/30s is a pretty generous limit, if someone hits that and they cant figure out that they were spamming then i dont know what to say lol 

Yeah that's what I was thinking also, the limit also appears to be changed to 40 commands every 60 seconds (according to a message by stupid cat) so it's even more generous. 

Should I just remove the mentions of the \*exact\* limit and only keep the mention of the timeout?

danny-may:
I think that makes sense!

HunteRoi (approved):

danny-may (approved):

PR #277, opened 2021-08-12
Add some undef checks for tags rendering

danny-may:
Not sure if this will fix all the issues, but hopefully will fix the current one. Cant reproduce it locally

### DIFF
--- a/src/backend/helpers.js
+++ b/src/backend/helpers.js
@@ -60,7 +60,7 @@ const tagType = {
 };
 
 function mdToHtml(text) {
-    if (text === undefined)
+    if (text == undefined)
         return undefined;
 
     const html = converter.makeHtml(text) || text;
@@ -68,7 +68,7 @@ function mdToHtml(text) {
 }
 
 function addSubtagReferences(text) {
-    if (text === undefined)
+    if (text == undefined)
         return undefined;
 
     return text.replace(/\{([a-z]+)\}/ig, function (match, subtag) {
--- a/src/backend/helpers.js
+++ b/src/backend/helpers.js
@@ -60,16 +60,17 @@ const tagType = {
 };
 
 function mdToHtml(text) {
-    // text = text.replace(/([,;/])(?=\S)/g, '$1\u200b');
-    let result = converter.makeHtml(text).replace(/\n/g, '<br>');
+    if (text === undefined)
+        return undefined;
 
-    // if (result.startsWith('<p>'))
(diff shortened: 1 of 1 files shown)


src/backend/helpers.js
```diff
@@ -60,16 +60,17 @@ const tagType = {
...
-    // text = text.replace(/([,;/])(?=\S)/g, '$1\u200b');
-    let result = converter.makeHtml(text).replace(/\n/g, '<br>');
+    if (text === undefined)
+        return undefined;
 
-    // if (result.startsWith('<p>'))
-    //     result = result.substr(3, result.length - 7);
-
-    return result;
+    const html = converter.makeHtml(text) || text;
```
HunteRoi, line 66: the `|| text` part seems a bit security-unwise to me. Is this mdToHtml method publicly accessible in any manner?

src/backend/helpers.js
```diff
@@ -60,16 +60,17 @@ const tagType = {
 };
 
 function mdToHtml(text) {
-    // text = text.replace(/([,;/])(?=\S)/g, '$1\u200b');
-    let result = converter.makeHtml(text).replace(/\n/g, '<br>');
+    if (text === undefined)
```
HunteRoi, line 63: If `text` is null, we could still have issues, right?

src/backend/helpers.js
```diff
@@ -60,16 +60,17 @@ const tagType = {
...
-    // if (result.startsWith('<p>'))
-    //     result = result.substr(3, result.length - 7);
-
-    return result;
+    const html = converter.makeHtml(text) || text;
+    return html.replace(/\n/g, '<br>');
 }
 
 function addSubtagReferences(text) {
+    if (text === undefined)
```
HunteRoi, line 71: Same: `text` could be null

HunteRoi (changes requested):

src/backend/helpers.js
danny-may: Nope, all this code is server sided.

src/backend/helpers.js
danny-may: Yep, but the error is specifically saying its undefined. Ive changed both checks to be nullish checks instead

src/backend/helpers.js
danny-may: Same as above

HunteRoi (approved):

PR #278, opened 2021-08-12
Add context overrides to apply

RagingLink:
This PR was at first to allow the use of `{apply}` on `{function}`s, but this also happens to fix issues where users could use `{apply}` to 'bypass' context.overrides:
- `{apply}` can be used inside `{lock}` to nest them
- `{apply}` can be used inside `{filter}`/`{waitreaction}`/`{waitmessage}` to use disabled subtags.

This PR fixes these issues, and adds as mentioned before `func.name` functionality.
This also allows `apply` to execute subtags like `{reactuser}`/`{reaction}`/`{params}`/`{paramsarray}`/`{paramslength}`.

### DIFF
--- a/src/tags/apply.js
+++ b/src/tags/apply.js
@@ -2,7 +2,7 @@
  * @Author: stupid cat
  * @Date: 2017-05-07 18:25:58
  * @Last Modified by: RagingLink
- * @Last Modified time: 2021-06-15 19:23:19
+ * @Last Modified time: 2021-08-13 00:37:30
  *
  * This project uses the AGPLv3 license. Please read the license file before using/adapting any of the code.
  */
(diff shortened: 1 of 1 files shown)


src/tags/apply.js
```diff
@@ -12,19 +12,28 @@ const Builder = require('../structures/TagBuilder'),
...
-            if (definition == null)
+            let runSubtag;
+            const name = args[0].toLowerCase();
+            if (context.state.overrides.hasOwnProperty(name)) {
+                runSubtag = context.state.overrides[name];
+            } else {
+                const tagDefinition = TagManager.get(name) || {};
+                runSubtag = context.state.overrides[tagDefinition.name] || tagDefinition.execute;
+            }
+            if (runSubtag === undefined)
```
danny-may, line 33: Change this to a nullish check, just to be safe

RagingLink: Yeah I looked at the code a bit and it *should* only be `undefined` or the actual override, but I guess it won't hurt to be safe.

danny-may (approved):

HunteRoi (approved):

PR #281, opened 2021-08-22
Allow ignoring of webhooks if a valid webhook ID is provided

RagingLink:
Added support for ignoring webhooks in `b!log`. This silently fails if blarg doesn't have the `Manage Webhooks` permission.

**Added:**
- Additional check for webhooks in `b!log`

image
The initial suggestion mentioned including webhooks in user parsing (like bu.getUser), but I fear that might break stuff that relies on user-specific fields. So I added basic support for `b!log` only.

### DIFF
--- a/src/dcommands/log.js
+++ b/src/dcommands/log.js
@@ -32,7 +32,7 @@ class LogCommand extends BaseCommand {
                 '\n- all - enables all of the events' +
                 '\n\n`ignore` adds a list of users to ignore from logging. Useful for ignoring bots.' +
                 '\n`track` removes users from the ignore list' +
-                '\n**Note:** When a webhook is provided for `users`, I must the `Manage Webhooks` permission to recognize them!'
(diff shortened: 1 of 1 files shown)


danny-may (approved):

src/dcommands/log.js
```diff
@@ -31,7 +31,8 @@ class LogCommand extends BaseCommand {
                 '\n- role:<id> - when a role gets added or removed' +
                 '\n- all - enables all of the events' +
                 '\n\n`ignore` adds a list of users to ignore from logging. Useful for ignoring bots.' +
-                '\n`track` removes users from the ignore list'
+                '\n`track` removes users from the ignore list' +
+                '\n**Note:** When a webhook is provided for `users`, I must the `Manage Webhooks` permission to recognize them!'
```
HunteRoi, line 35: you missed the word "have" there, buddy

HunteRoi (changes requested):

HunteRoi (approved):

PR #284, opened 2021-08-30
Changing architecture of kick & ban dcommand and tag

toslaw1:
~~This should (I hope) fix void not supressing kick errors by changing the way they're handled from old switch to something similar to how ban subtag's errors are handled.~~

This pull request is supposed to change the architecture of ban and kick tag and default command. Both of them used completely different logic making it confusing. They are now both made in the better way (taken from ban tag and dcommand). This PR also contains other improvements, e. g. adding better replies to the commands.

**Change Log**
- Changed variable initialization system to use newer `let` instead of old `var`
- Changed variable names in kick dcommand & tag to make them same as in ban dcommand and tag
- Changed return types & values in kick command replacing numbers reprising the state with strings explaining the state
- Changed ban and kick commands' replies adding reason to them
- Added the code responsible for checking whether the user is on the guild or not to kick command fixing an issue causing an internal server error
- Replaced `moment#humanize` with discord timestamps

### DIFF
--- a/src/dcommands/kick.js
+++ b/src/dcommands/kick.js
@@ -21,12 +21,12 @@ class KickCommand extends BaseCommand {
             }
             if (!context.guild.members.get(user.id)) return this.userNotInGuild(subtag, context);   //checking if user is in guild; if not then returning user not in guild error
             
-            let target = await bu.getUser(msg, words[1]);
+            let target = await bu.getUser(msg, words[1]);     //unused
             let reason = bu.parseInput(this.flags, words).r;
 
             if (!target) return;
 
-            let state = await this.kick(msg, target, reason, false, false);
+            /*let state = await this.kick(msg, target, reason, false, false);
             let response;
             switch (state) {
                 case 0: //Successful
@@ -47,27 +47,37 @@ class KickCommand extends BaseCommand {
                 default: //Error occurred
                     response = `Failed to kick the user! Please check your permission settings and command and retry. \nIf you still can't get it to work, please report it to me by doing \`b!report <your issue>\` with the following:\`\`\`\n${state.message}\n${state.response}\`\`\``;
                     break;
-        }
+            }
 
-        bu.send(msg, response);
+            bu.send(msg, response);*/
+            
+            bu.send(await this.kick(msg, user, reason));
             
         } else bu.send(msg, `You didn't tell me who to kick!`);
     }
 
-    async kick(msg, target, reason, tag = false, noPerms = false) {
-        if (!msg.channel.guild.members.get(bot.user.id).permissions.json.kickMembers)
-            return 1;
+    async kick(msg, user, reason, tag = false, noPerms = false) {
+        if (!msg.channel.guild.members.get(bot.user.id).permissions.json.kickMembers) {
+            return [`I don't have permission to kick users!`, '`Bot has no permissions`'];
+        }
         let kickPerms = await bu.guildSettings.get(msg.guild.id, 'kickoverride') || 0;
-        if (!noPerms && !bu.comparePerms(msg.member, kickPerms) && !msg.member.permissions.json.kickMembers)
-            return 3;
-
-        var botPos = bu.getPosition(msg.channel.guild.members.get(bot.user.id));
-        var userPos = bu.getPosition(msg.member);
-        var targetPos = bu.getPosition(msg.channel.guild.members.get(target.id));
-        if (targetPos >= botPos)
-            return 2;
-        if (!noPerms && targetPos >= userPos && msg.author.id != msg.guild.ownerID)
-            return 4;
+        if (!noPerms && !bu.comparePerms(msg.member, kickPerms) && !msg.member.permissions.json.kickMembers) {
+            return [`You don't have permission to ban users!`, '`User has no permissions`'];
+        }
+        
+        let member = msg.guild.members.get(user.id);
+        
+        if(member) {
+            var botPos = bu.getPosition(msg.channel.guild.members.get(bot.user.id));
+            var userPos = bu.getPosition(msg.member);
+            var targetPos = bu.getPosition(msg.channel.guild.members.get(user.id));
+            if (targetPos >= botPos) {
+                return [`I don't have permission to kick ${user.username}!`, '`Bot has no permissions`'];
+            }
+            if (!noPerms && targetPos >= userPos && msg.author.id != msg.guild.ownerID) {
+                return [`You don't have permission to kick ${user.username}!`, '`User has no permissions`'];
+            }
+        }
 
         try {
             const fullReason = encodeURIComponent((tag ? '' : `[ ${bu.getFullName(msg.author)} ]`)
@@ -79,11 +89,11 @@ class KickCommand extends BaseCommand {
                 fullReason
             );
 
-            return 0;
-        }
-        catch (err) {
-            console.error(err);
-            return err;
+            return [`:ok_hand: Kicked ${user.username}. Reason: ${reason}`];
+        } catch (err) {
+            //console.error(err);
+            //return err;
+            return [`Failed to kick the user! Please check your permission settings and command and retry. \nIf you still can't get it to work, please report it to me by doing \`b!report <your issue>\` with the following:\`\`\`\n${err.message}\n${err.response}\`\`\``, false];
         }
     }
 }
--- a/src/dcommands/kick.js
+++ b/src/dcommands/kick.js
@@ -12,40 +12,46 @@ class KickCommand extends BaseCommand {
     }
 
     async execute(msg, words, text) {
-        if (!words[1]) {
-            bu.send(msg, `You didn't tell me who to kick!`);
-            return;
-        }
+        if (words[1]) {
+            let input = bu.parseInput(this.flags, words);
+            
+            var user = await bu.getUser(msg, input.undefined[0]);
+            if (!user) {
+                return await bu.send(msg, `I couldn't find that user. Try again with their ID or a mention instead.`);
+            }
+            if (!context.guild.members.get(user.id)) return this.userNotInGuild(subtag, context);   //checking if user is in guild; if not then returning user not in guild error
+            
+            let target = await bu.getUser(msg, words[1]);
+            let reason = bu.parseInput(this.flags, words).r;
 
-        let target = await bu.getUser(msg, words[1]);
-        let reason = bu.parseInput(this.flags, words).r;
+            if (!target) return;
 
-        if (!target) return;
-
-        let state = await this.kick(msg, target, reason, false, false);
-        let response;
-        switch (state) {
-            case 0: //Successful
-                response = `:ok_hand:`;
+            let state = await this.kick(msg, target, reason, false, false);
+            let response;
+            switch (state) {
+                case 0: //Successful
+                    response = `Kicked ${target.username}. Reason: ${reason}`;
+                    break;
+                case 1: //Bot doesnt have perms
+                    response = `I don't have permission to kick users!`;
+                    break;
+                case 2: //Bot cannot kick target
+                    response = `I don't have permission to kick ${target.username}!`;
                 break;
-            case 1: //Bot doesnt have perms
-                response = `I don't have permission to kick users!`;
-                break;
-            case 2: //Bot cannot kick target
-                response = `I don't have permission to kick ${target.username}!`;
-                break;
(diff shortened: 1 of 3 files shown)


src/tags/kick.js
```diff
@@ -42,21 +42,9 @@ module.exports =
...
-                    return error(`I don't have permission to kick ${user.username}!`);
-                case 3: //User doesnt have perms
-                    return error(`You don't have permission to kick users!`);
-                case 4: //User cannot kick target
-                    return error(`You don't have permission to kick ${user.username}!`);
-                default: //Error occurred
-                    throw state;
-            }
+            if (typeof response[1] == 'string' && response[1].startsWith('`'))
+            return Builder.util.error(subtag, context, response[1]);
```
HunteRoi, line 46: This line needs to be indented once again so that the return is part of the if scope and that we can access the line 47 in case we have no error.

HunteRoi (changes requested):
Hey there!

First of all, thank you for your contribution. This indeed looks like the way to go. I commented herebelow the line 46 which should be tabbed once to the right so that it falls under the `if` block and does not prevent from accessing the return in line 47.

Also, I'm going to request @Titansmasher to review this because I'm not well trained on the current underlying process of the BBTag Engine. Furthermore, I am a little bit concerned about [the lines 172 to 173 of the `Builder.util.error` method](https://github.com/blargbot/blargbot/blob/72f8151f81c5abcae306bad89908e10d51395e91/src/structures/bbtag/Engine.js#L172:#L173), because it seems like the condition is wrong for most of the use cases I have in mind.

Bear with us a little more so that we can make sure your PR is on tracks for the next step!
Once again, thank you for your contribution. We all look forward to see more of your work here 😄

src/tags/kick.js
danny-may: Indents dont matter for js, but yes this should be indented for style reasons

danny-may (changes requested):
A couple of points:
1. After [discussing the issue with ASoD](https://discord.com/channels/194232473931087872/229137183234064384/881859558652399666) We worked out that the error was unrelated to the kick subtag, it was coming from the dm subtag instead.
2. I think this is still a good change to make from an architecture point of view, but there will need to also be changes to the [kick command] too. The ban subtag depends on the ban command, and that is structured differently to how the kick subtag and command work. We would need to update the kick command to match the strucutre of the ban command.
3. If you want to make those changes still, adding some proper responses to the kick and ban commands would be pretty rad rather than just 👌

RagingLink:
Perhaps it'd be nice to add a check to see if the user is in the current guild, if you want to change the kick errors that is, this would (hopefully) prevent `An internal server error has occurred`. 
Think it'd be something like `if (!context.guild.members.get(user.id)) return this.userNotInGuild(subtag, context)`

toslaw1:
Hey guys, thank you for your comments. I'll try to look at kick & ban command and update what is required. If I'll need any reviews or something like that I'll be sure to ask you.

toslaw1:
Not sure if my changes make any sense, hope they do lol, I'm a noob

toslaw1:
Alright, I think it should be ready. I've got one more question. Is there any reason for using `var` except `let` for some variables in `ban.js`? I followed the style from `ban.js` but if there is no reason then I might change that to `let` in both files as IIRC `var` is not recommended.

toslaw1:
Didn't notice that I didn't change style of kick tag to ban's one, should be fixed now. You can review it now to get sure I did it right and didn't miss something.

danny-may:
> 
> 
> Alright, I think it should be ready. I've got one more question. Is there any reason for using `var` except `let` for some variables in `ban.js`? I followed the style from `ban.js` but if there is no reason then I might change that to `let` in both files as IIRC `var` is not recommended.

let > var. Most linting now will tell you to never use var. If the value of something isnt going to change, use `const`, otherwise use `let`

toslaw1:
Okay then I'll change var to let in both kick and ban, just give me a while.

toslaw1:
Okay, did some changes, waiting for comments.

src/dcommands/kick.js
```diff
@@ -12,56 +12,72 @@ class KickCommand extends BaseCommand {
...
-
-        var botPos = bu.getPosition(msg.channel.guild.members.get(bot.user.id));
-        var userPos = bu.getPosition(msg.member);
-        var targetPos = bu.getPosition(msg.channel.guild.members.get(target.id));
-        if (targetPos >= botPos)
-            return 2;
-        if (!noPerms && targetPos >= userPos && msg.author.id != msg.guild.ownerID)
-            return 4;
+        if (!noPerms && !bu.comparePerms(msg.member, kickPerms) && !msg.member.permissions.json.kickMembers) {
+            return [`You don't have permission to ban users!`, '`User has no permissions`'];
```
danny-may, line 65: Should say `kick users`

HunteRoi:
Can you update the PR title and the description to properly describe what you are indenting to do with the changes? A good way of doing this is by listing all changes.

_e.g:_
This PR aims to improve the overall architecture of the application concerning the behavior of kick and ban. These two actions are activated differently whether you trigger them with the default command or with the subtag. The idea is to use the same underlying logic in order to have the same behavior, and merge similar structures of code together to prevent inconsistency and bugs.

**Changed**
- initialize variables with `let` instead `var` in the `ban` and `kick` dcommands & subtags
- use Discord Timestamps instead of `moment#humanize`
- ...

src/dcommands/kick.js
```diff
@@ -12,56 +12,72 @@ class KickCommand extends BaseCommand {
...
+                    break;
+                default: //Error occurred
+                    response = `Failed to kick the user! Please check your permission settings and command and retry. \nIf you still can't get it to work, please report it to me by doing \`b!report <your issue>\` with the following:\`\`\`\n${state.message}\n${state.response}\`\`\``;
+                    break;
+            }
 
-        bu.send(msg, response);
+            bu.send(msg, response);*/
+            
+            bu.send(await this.kick(msg, user, reason));
```
danny-may, line 54: `this.kick(...)` now returns `Promise<[string, string]>` so you need to access the 0th index after awaiting
```suggestion
            bu.send((await this.kick(msg, user, reason))[0]);

```

src/dcommands/kick.js
```diff
@@ -73,11 +89,11 @@ class KickCommand extends BaseCommand {
...
             );
 
-            return 0;
-        }
-        catch (err) {
-            console.error(err);
-            return err;
+            return [`:ok_hand: Kicked ${user.username}. Reason: ${reason}`];
+        } catch (err) {
+            //console.error(err);
```
danny-may, line 94: Leave the `console.error` uncommented

src/dcommands/kick.js
```diff
@@ -12,56 +12,72 @@ class KickCommand extends BaseCommand {
...
+            let botPos = bu.getPosition(msg.channel.guild.members.get(bot.user.id));
+            let userPos = bu.getPosition(msg.member);
+            let targetPos = bu.getPosition(msg.channel.guild.members.get(user.id));
+            if (targetPos >= botPos) {
+                return [`I don't have permission to kick ${user.username}!`, '`Bot has no permissions`'];
+            }
+            if (!noPerms && targetPos >= userPos && msg.author.id != msg.guild.ownerID) {
+                return [`You don't have permission to kick ${user.username}!`, '`User has no permissions`'];
+            }
+        }
```
danny-may, line 80: I would change this. You can ban someone who isnt on the guild, but not kick them. This block in the ban command is to check if you are able to ban them if they are on the guild. I think just erroring if they arent is a better idea for kick rather than getting an error thrown below.
```suggestion
        let member = msg.guild.members.get(user.id);
        
        if(!member) {
            return [`${user.username} isnt on this server!`, '`User not in guild`'];
        }
        let botPos = bu.getPosition(msg.channel.guild.members.get(bot.user.id));
        let userPos = bu.getPosition(msg.member);
        let targetPos = bu.getPosition(msg.channel.guild.members.get(user.id));
        if (targetPos >= botPos) {
            return [`I don't have permission to kick ${user.username}!`, '`Bot has no permissions`'];
        }
        if (!noPerms && targetPos >= userPos && msg.author.id != msg.guild.ownerID) {
            return [`You don't have permission to kick ${user.username}!`, '`User has no permissions`'];
        }

```

danny-may (changes requested):

toslaw1:
Thanks guys. I'll change the title and description and make requested changes when I'll be home.

toslaw1:
I applied Danny's suggestions and I also changed the PR title, however I don't know how to change the PR description. Does anything like that exist? Thanks for help

toslaw1:
I updated the description. I'm also wondering if the `bug` label shouldn't be removed as the purpose of the PR changed

src/dcommands/kick.js
```diff
@@ -12,56 +12,73 @@ class KickCommand extends BaseCommand {
...
+        
+        let member = msg.guild.members.get(user.id);
+        
+        if(!member) {
+            return [`${user.username} isnt on this server!`, '`User not in guild`'];
+        }
+        let botPos = bu.getPosition(msg.channel.guild.members.get(bot.user.id));
+        let userPos = bu.getPosition(msg.member);
+        let targetPos = bu.getPosition(msg.channel.guild.members.get(user.id));
```
HunteRoi, line 75: If these are not updated anywhere, you can declare them as const

src/dcommands/kick.js
```diff
@@ -12,56 +12,73 @@ class KickCommand extends BaseCommand {
...
-        var targetPos = bu.getPosition(msg.channel.guild.members.get(target.id));
-        if (targetPos >= botPos)
-            return 2;
-        if (!noPerms && targetPos >= userPos && msg.author.id != msg.guild.ownerID)
-            return 4;
+        if (!noPerms && !bu.comparePerms(msg.member, kickPerms) && !msg.member.permissions.json.kickMembers) {
+            return [`You don't have permission to kick users!`, '`User has no permissions`'];
+        }
+        
+        let member = msg.guild.members.get(user.id);
```
HunteRoi, line 68: Never updated, could be declared as const

src/dcommands/kick.js
HunteRoi, line 75: The `msg.channel.guild.members.get(user.id)` part is already done when declaring `member` variable. You can thus use it here instead of searching for the member again

HunteRoi (changes requested):

src/tags/kick.js
```diff
@@ -56,7 +56,12 @@ module.exports =
                     return error(`You don't have permission to kick ${user.username}!`);
                 default: //Error occurred
                     throw state;
-            }
+            }*/
+    
+            if (typeof response[1] == 'string' && response[1].startsWith('`'))
+                return Builder.util.error(subtag, context, response[1]);
+    
+            return response[1];
```
danny-may, line 64: Check and throw if there is an error
```suggestion
            if (typeof response[1] != 'string')
                throw response[1];
            if (response[1].startsWith('`'))
                return Builder.util.error(subtag, context, response[1].slice(1, -1));
    
            return response[1];
```

src/dcommands/kick.js
```diff
@@ -12,56 +12,73 @@ class KickCommand extends BaseCommand {
...
+                    break;
+                default: //Error occurred
+                    response = `Failed to kick the user! Please check your permission settings and command and retry. \nIf you still can't get it to work, please report it to me by doing \`b!report <your issue>\` with the following:\`\`\`\n${state.message}\n${state.response}\`\`\``;
+                    break;
+            }
 
-        bu.send(msg, response);
+            bu.send(msg, response);*/
+            
+            bu.send((await this.kick(msg, user, reason))[0]);
```
danny-may, line 54: Missing a context argument for send
```suggestion
            bu.send(msg, (await this.kick(msg, user, reason))[0]);
```

src/tags/kick.js
```diff
@@ -56,7 +56,12 @@ module.exports =
                     return error(`You don't have permission to kick ${user.username}!`);
                 default: //Error occurred
                     throw state;
-            }
+            }*/
```
danny-may, line 59: Delete this, no need to comment code

danny-may (changes requested):
Just a few changes. Ive tested this locally and with the changes it looks fine to me!

toslaw1:
Alright, thanks, Danny. Done I think.

danny-may (approved):
LGTM, just need to fix the merge conflict!

PR #288, opened 2021-09-11
Added more attributes to the xss whitelist

RagingLink:
Added `style`, `title` and `data-tooltip` to the whitelisted attributes for the output page. I went through the attribute list to see if there are perhaps more attributes blarg could support, but I didn't really find anything 'worth' adding.

**Added:**
- `style`, `title` and `data-tooltip` to the global attribute whitelist.

Stems from a feature request in `#bb-support`:
image

### DIFF
--- a/src/backend/routes/output.js
+++ b/src/backend/routes/output.js
@@ -2,7 +2,7 @@
  * @Author: stupid cat
  * @Date: 2017-05-07 18:19:10
  * @Last Modified by: RagingLink
- * @Last Modified time: 2021-09-11 12:52:57
+ * @Last Modified time: 2021-09-11 12:58:55
  *
  * This project uses the AGPLv3 license. Please read the license file before using/adapting any of the code.
  */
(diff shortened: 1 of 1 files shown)


HunteRoi:
Watch out with `data-` attributes, they could be used by javascript libraries and could lead to flaws in the security of the bot.

RagingLink:
> Watch out with `data-` attributes, they could be used by javascript libraries and could lead to flaws in the security of the bot.

Can't quite say I understand where you're coming from tbh.

~~I was typing out some examples to explain my view, but I honestly cannot see how it'd impact the security server-side~~

danny-may (approved):

HunteRoi (approved):
My previous comment was just a warning. Not all `data-` attribute should be whitelisted.

But this list is fine! Thank you :smile:

RagingLink:
> 
> 
> My previous comment was just a warning. Not all `data-` attribute should be whitelisted.
> 
> But this list is fine! Thank you 😄

Ah yeah that's understandable o/ Your previous comment confused me a bit as `data-tooltip` shouldn't be an issue

PR #290, opened 2021-09-17
Improve lookup mechanism to respect exact matches

danny-may:
Add support to return exact matches preferentially for user/channel/role lookups

Requested [here](https://discord.com/channels/194232473931087872/238131984017260546/887723443435499530)


.vscode/launch.json
```diff
@@ -0,0 +1,19 @@
...
+            "name": "Launch",
+            "skipFiles": [
+                "<node_internals>/**"
+            ],
+            "autoAttachChildProcesses": true,
+            "console": "integratedTerminal",
+            "program": "${workspaceFolder}/index.js"
+        }
+    ]
+}
```
HunteRoi, line 19: I'm not sure this is needed for the change request. Is it possible that you remove it from your pr ?

HunteRoi (approved):

.vscode/launch.json
danny-may: But I wanted to be able to just press F5 to run the bot ;-;

Point taken though, its not needed here I guess

PR #291, opened 2021-09-17
Dont keep looping after censor limit is hit

### LINKED ISSUE #289: [BUG] Weird behaviour on long censorlist
How to reproduce:

- Have more than 77 censors (tested with 154)
- run `b!censor info`

What is expected to happen:
- The bot gives the list of existing censors

What happened instead:
- The bot posted:
```
Oops! I tried to send a message that was too long. If you think this is a bug, please report it!

To see what I would have said, please visit [link to output]
```

- The output in the output link is truncated at 77 and then starts repeating:
```
...and 77 more.
...and 76 more.
...and 75 more.
...and 74 more.
...and 73 more.
...and 72 more.
...and 71 more.
...and 70 more.
...and 69 more.
...and 68 more.
...and 67 more.
...and 66 more.
...and 65 more.
...and 64 more.
...and 63 more.
```

danny-may:
fixes #289

### DIFF
--- a/src/dcommands/censor.js
+++ b/src/dcommands/censor.js
@@ -353,6 +353,7 @@ class CensorCommand extends BaseCommand {
                     let phrase = `${i + 1}. ${storedGuild.censor.list[i].term}${storedGuild.censor.list[i].regex ? ' (regex)' : ''}\n`;
                     if (censorList.length + phrase.length + suffix.length > 1500) {
                         censorList += `...and ${storedGuild.censor.list.length - i} more.\n`;
+                        break;
                     } else {
                         censorList += phrase;
                     }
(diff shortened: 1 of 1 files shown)


RagingLink:
Is there any real reason for the limit of `1500` characters? IMO the limit could just be removed

HunteRoi (approved):

HunteRoi:
The bot can go up to 4000 and also has something to auto-split and send long messages now, right ?

RagingLink:
> The bot can go up to 4000 and also has something to auto-split and send long messages now, right ? 

Yeah if we still want to limit the output to fit in a discord message it could be increased to 4000 (or 3500 if the 500 characters are used for anything else). I'm not sure if blarg  auto-splits messages, but as there is an output page for long messages the limit could just be removed.

danny-may:
We could remove the limit and rely on the output page to be generated, but I feel like that can be a different PR (Or just leave it for the TS rewrite)

RagingLink:
Yeah that's fine

RagingLink (approved):

PR #292, opened 2021-09-17
Typescript rewrite

danny-may:
This is a rewrite of the entire bot to typescript.

Theres too many changes to list them all, but the main goal of this is to get type safety added to the codebase without losing any functionality, and without breaking any subtags.


HunteRoi:
image

danny-may:
Only 326? Those are rookie numbers

PR #304, opened 2021-11-02
Switch to the rapidapi host rather than mashape

danny-may:
The words API moved hosts from mashape to rapidapi a long time ago, but blarg never updated.
New docs: https://rapidapi.com/dpventures/api/wordsapi/

### DIFF
--- a/src/dcommands/define.js
+++ b/src/dcommands/define.js
@@ -41,9 +41,10 @@ class DefineCommand extends BaseCommand {
         vars.uses++;
         await r.table('vars').get('wordapis').update(vars);
         request({
-            url: `https://wordsapiv1.p.mashape.com/words/${args}`,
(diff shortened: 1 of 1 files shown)


RagingLink (approved):

HunteRoi (approved):
This will require to add the RapidAPI apikey into the config of the production bot when we wil update the process.

danny-may:
> 
> 
> This will require to add the RapidAPI apikey into the config of the production bot when we wil update the process.

The mashape key is already in the config, this is just a host change. The old key should work for rapidapi I believe

HunteRoi:
> > This will require to add the RapidAPI apikey into the config of the production bot when we wil update the process.
> 
> The mashape key is already in the config, this is just a host change. The old key should work for rapidapi I believe

Great news! Thank you for looking into that and update the file! :smile:

PR #323, opened 2022-03-05
VersionStateManager

HunteRoi:
- Adds a `VersionStateManager` to encapsulate the logic of retrieving, parsing and updating the version from/to the database.
- Updates the related default commands to use it as well.

### DIFF
--- a/src/workers/cluster/managers/VersionStateManager.ts
+++ b/src/workers/cluster/managers/VersionStateManager.ts
@@ -1,0 +1,44 @@
+import { mapping } from '@cluster/utils';
+import { VarsTable } from '@core/types';
+import { UpdateType, Version, VersionManager } from '@hunteroi/versioning';
+
+const mapUpdateType = mapping.choice(
+    mapping.in<UpdateType>('major', 'minor'),
+    mapping.string.map<UpdateType>(() => 'patch')
+);
+
+export class VersionStateManager {
+    #version: Version;
+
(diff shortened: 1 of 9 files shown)


src/workers/cluster/managers/VersionStateManager.ts
```diff
@@ -0,0 +1,44 @@
...
+        if (!mapped.valid) {
+            throw new Error('Invalid update type');
+        }
+
+        manager.update(mapped.value);
+
+        await this.db.set('version', this.#version);
+    }
+
+    private async refresh(): Promise<void> {
```
danny-may, line 36: Dont really need this refresh and cache pattern I think. Move this to `getVersion` and make `updateVersion` call `getVersion`. Can get rid of the `#version` field then

danny-may (changes requested):

src/workers/cluster/managers/VersionStateManager.ts
HunteRoi: Then `getVersion` would return a `Version` instance instead of a string. But we don't need that outside the manager as we always display the version as a string. I could just change the `refresh` to something like `getFromStorage` and it would return the `Version` instance directly instead of assigning it to the private field.

What do you think?

danny-may: Yep youre right. `getFromStorage` sounds better!

danny-may (approved):

PR #338, opened 2022-06-11
Timeout support

HunteRoi:
**Added**
- a `timeout` command to moderate users
- support in event-log and modlog for timeouts
- `{timeout}` and `{usertimedout}` subtags to moderate a user or get the date of their timeout
- censors now handles timing out users
- settings have now a `timeoutat` key

**Changed**
- the `ban` and `mute` commands now support the same pattern as `timeout` does, for unban and unmute features
- fix a bug where the moderator would never be set on eventlog
- some docs were adjusted in the process (eg.: kick subtag, userisbot subtag, ban command, mute command)

Note: if someone needs to change the timeout duration on censors, they will have to handle timing out users on their own via the `{timeout}` subtag. The default value is set to 1 day on purpose and I do not plan to support a setting to set a timeout's duration.


src/bbtag/subtags/user/usertimeout.ts
```diff
@@ -0,0 +1,61 @@
+import { Member } from 'eris';
+import moment from 'moment-timezone';
+
+import { BBTagContext } from '../../BBTagContext';
+import { CompiledSubtag } from '../../compilation';
+import { BBTagRuntimeError, UserNotFoundError } from '../../errors';
+import { SubtagType } from '../../utils';
+
+export class UserTimeOutSubtag extends CompiledSubtag {
```
danny-may, line 9: `UserTimeoutSubtag`

src/cluster/managers/moderation/EventLogManager.ts
```diff
@@ -9,16 +9,46 @@ export class EventLogManager {
...
+    }
+
+    public async userUnTimedOut(member: Member): Promise<void> {
+        const channel = await this.getLogChannel('memberuntimeout', member.guild.id);
+        if (channel === undefined || await this.isExempt(member.guild.id, member.id))
+            return;
+
+        const now = moment();
+        const auditEvents = await tryGetAuditLogs(member.guild, 50, undefined, AuditLogActionType.MEMBER_UPDATE);
+        const audit = auditEvents?.entries.find(e => e.targetID === member.id && moment(e.createdAt).isAfter(now.add(-1, 'second')));
```
danny-may, line 30: This might need to be revisited, the user might have been updated multiple times I guess. We should look specifically for one where the timeout was updated

src/cluster/managers/moderation/TimeoutManager.ts
```diff
@@ -0,0 +1,90 @@
...
+            return permMessage;
+
+        this.ignoreUnTimeouts.add(`${guild.id}:${member.id}`);
+        await guild.editMember(member.id, { communicationDisabledUntil: null }, `[${humanize.fullName(moderator)}] ${reason ?? ''}`);
+        await this.modLog.logUnTimeout(guild, member.user, moderator, reason);
+
+        return 'success';
+    }
+
+    private async tryTimeoutUser(guild: Guild, userId: string, moderator: User, authorizer: User, duration: Duration, reason?: string): Promise<TimeoutResult | { error: unknown; }> {
```
danny-may, line 54: Not an issue if we leave it like this, but the try pattern isnt needed for timeouts. Its there for bans because massban is a thing. Up to you if you leave it here

danny-may (changes requested):

src/bbtag/subtags/user/timeout.ts
```diff
@@ -0,0 +1,82 @@
+import { clampBy, parse } from '@blargbot/core/utils';
+import moment from 'moment-timezone';
+
+import { BBTagContext } from '../../BBTagContext';
+import { CompiledSubtag } from '../../compilation';
+import { BBTagRuntimeError, UserNotFoundError } from '../../errors';
+import { SubtagType } from '../../utils';
+
+const maximumTimeoutDurationInSeconds = 2419190; // 28 days - 10s because Discord throws a RESTError when the duration is too close to 28d
```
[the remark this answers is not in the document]
HunteRoi: I tried but it sucks for the tests. I cannot be sure that we clamped the duration if the clamping happens in the manager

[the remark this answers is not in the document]
danny-may: Thats fine, the tests should just be asserting the manager is called with the correct args. We can add tests that the manager handles the args correctly separately, but I dont think we need to do that right now. Key thing for the tests is ensuring the interface is tested, so we wont accidentally break peoples code with updates. Things like the date being clamped is an implementation detail that doesnt affect the subtag interface imo.

src/cluster/managers/moderation/EventLogManager.ts
HunteRoi: That means I need to add a check in the find condition to be sure the communicationDisabledUntil property is updated, right?

danny-may: Might be? Im not sure, we havent drilled any deeper into an audit log before besides its type. Im not sure how the details of the change are represented in there

src/cluster/managers/moderation/TimeoutManager.ts
HunteRoi: I actually find it pretty useful to separate the actual action into a private method. We don't need to overcomplicate the public `timeout` method. I think I will let the method but rename it.

Should I keep the ignoreTimeouts and ignoreTimeoutClears though?

danny-may (approved):

RagingLink (changes requested):
I noticed that the timeout subtags weren't present at all when testing them, this seems to be due to them not being in `src/bbtag/subtags/user/index.ts` (can be generated with `src/bbtag/subtags/_generateIndexFiles.ts`, but it also seems to be a too big of an oversight lol so it might be me).

 I also wonder why you are against adding a config for the duration of timeouts given after being warned. As it seems like a trivial thing to add, and this doesn't make it necessary to make a custom warn command if you do want the customization

HunteRoi:
Thnak you for the feedbacks @danny-may @RagingLink ! Regarding the last review, here is my response:

> I noticed that the timeout subtags weren't present at all when testing them, this seems to be due to them not being in `src/bbtag/subtags/user/index.ts` (can be generated with `src/bbtag/subtags/_generateIndexFiles.ts`, but it also seems to be a too big of an oversight lol so it might be me).

You are right! I forgot that part, as I actually haven't tested them officially. I directly went on unit testing and forgot integration tests... I'll update that part! Did you test them on your local machine though?

> I also wonder why you are against adding a config for the duration of timeouts given after being warned. As it seems like a trivial thing to add, and this doesn't make it necessary to make a custom warn command if you do want the customization

I don't think it is actually necessary to allow people to set the duration through the default system. That is not "much" to do, but it still requires:
- to add a system to parse the duration from a user input
- to store and retrieve the duration in and from the guild settings 
- to decide what to do when the user gives a lower value than 0s or a higher value than 28d... (store as is, clamp the value and that's it ; send a message from the settings command to say the value is not valid ; ...) knowing that a clamping system is already applied in the timeout moderation manager.

If you think this would be a good addition, feel free to add it by editing this branch or with a new branch!

RagingLink (approved):
Your reasons for not having a custom timeout when being warned are understandable, it could always be added if it happens to be a requested feature :)

HunteRoi:
I mean yeah of course it could be added. So could a lot of things be added to the censor system...
When I talked about it with @danny-may, I remember he told me that even the kick and ban procedures could be avoided as they all could be done via BBTag on the message deletion. I agree with that. I would even be on the side of removing the support for those kick and ban procedures and tell people to use BBTag. That is why I would try to setup as less as possible for custom things and let people do their own! 😄

PR #339, opened 2022-06-17
Misc fixes

danny-may:
- Regression fix for {waitreact} and {waitmessage}.
  - if the condition doesnt return `true` or `false`, an error should be thrown. [Old code](https://github.com/blargbot/blargbot/blob/1189d139610f51e9fec1c19871f821fc166d238d/src/tags/waitmessage.js#L146-L150)

- Regression fix for {jget;~var;}
  - In js blarg this used to get the `""` property (empty key). 
  image

- Fix for define command not working
  - Api response structure added new props but the parser was strict so rejected the payload

### DIFF
--- a/src/bbtag/utils/json.ts
+++ b/src/bbtag/utils/json.ts
@@ -85,8 +85,6 @@ export const json = Object.freeze({
     getPathKeys(path: string | readonly string[]): readonly string[] {
         if (typeof path !== 'string')
             return path;
-        if (path === '')
-            return [];
         return path.split('.');
     },
     parse(value: string): JToken | undefined {
--- a/src/cluster/dcommands/general/define.ts
+++ b/src/cluster/dcommands/general/define.ts
@@ -74,7 +74,7 @@ const wordApiMapping = mapping.object({
         definition: mapping.string,
         partOfSpeech: mapping.string,
         synonyms: mapping.array(mapping.string).optional
-    })),
+    }, { strict: false })),
     frequency: mapping.number,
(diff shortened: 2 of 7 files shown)


src/bbtag/utils/json.ts
```diff
@@ -85,8 +85,6 @@ export const json = Object.freeze({
     getPathKeys(path: string | readonly string[]): readonly string[] {
         if (typeof path !== 'string')
             return path;
-        if (path === '')
-            return [];
```
HunteRoi, line 89: What happens when you do `getPathKeys("");` ?

src/cluster/dcommands/general/define.ts
```diff
@@ -74,7 +74,7 @@ const wordApiMapping = mapping.object({
         definition: mapping.string,
         partOfSpeech: mapping.string,
         synonyms: mapping.array(mapping.string).optional
-    })),
+    }, { strict: false })),
```
HunteRoi, line 77: What does it mean?

src/bbtag/utils/json.ts
danny-may: you get back `[""]` now, which means "go to the empty key"

src/cluster/dcommands/general/define.ts
danny-may: object mappings are strict by default which means that the source object cannot have any excess properties that you dont map. Making it not strict means that it only requires the properties you do define

HunteRoi (approved):

PR #344, opened 2022-06-18
Fix UTC+0 offset not being applied in {userboostdate}

RagingLink:
This offsets the moment to utc 0, just like it was on blarg-js.

This also fixes wrong example docs for `{userboosdate;<format>;<user>;[quiet]}` and fixes a classname typo

### DIFF
--- a/src/bbtag/subtags/user/userboostdate.ts
+++ b/src/bbtag/subtags/user/userboostdate.ts
@@ -6,7 +6,7 @@ import { CompiledSubtag } from '../../compilation';
 import { BBTagRuntimeError, UserNotFoundError } from '../../errors';
 import { SubtagType } from '../../utils';
 
-export class UserBoostDataSubtag extends CompiledSubtag {
+export class UserBoostDateSubtag extends CompiledSubtag {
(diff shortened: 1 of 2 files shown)


danny-may (approved):

src/bbtag/subtags/user/userboostdate.ts
```diff
@@ -6,7 +6,7 @@ import { CompiledSubtag } from '../../compilation';
 import { BBTagRuntimeError, UserNotFoundError } from '../../errors';
 import { SubtagType } from '../../utils';
 
-export class UserBoostDataSubtag extends CompiledSubtag {
+export class UserBoostDateSubtag extends CompiledSubtag {
```
HunteRoi, line 9: Why? I don't even see the change here 😂

HunteRoi (approved):

src/bbtag/subtags/user/userboostdate.ts
RagingLink: UserBoost ***Data*** vs UserBoost ***Date*** 🙂

PR #345, opened 2022-06-18
Fix some sentry errors

danny-may:
- Fixes sentry logging errors with title `sendToSentry`. Now has a more meaningful error
- Fixes [#BLARGBOT-375](https://sentry.blargbot.xyz/organizations/blargbot/issues/3313)
- Fixes [#BLARGBOT-37B](https://sentry.blargbot.xyz/organizations/blargbot/issues/3319)

### DIFF
--- a/src/cluster/managers/moderation/ChatLogManager.ts
+++ b/src/cluster/managers/moderation/ChatLogManager.ts
@@ -52,7 +52,7 @@ export class ChatLogManager {
     }
 
     public async messageUpdated(message: Message<PossiblyUncachedTextableChannel>): Promise<void> {
-        if (!guard.isGuildMessage(message) || await this.cluster.database.guilds.getSetting(message.channel.guild.id, 'makelogs') !== true)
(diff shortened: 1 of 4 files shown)


src/cluster/managers/moderation/ChatLogManager.ts
```diff
@@ -52,7 +52,7 @@ export class ChatLogManager {
     }
 
     public async messageUpdated(message: Message<PossiblyUncachedTextableChannel>): Promise<void> {
-        if (!guard.isGuildMessage(message) || await this.cluster.database.guilds.getSetting(message.channel.guild.id, 'makelogs') !== true)
+        if (!guard.isGuildMessage(message) || await this.cluster.database.guilds.getSetting(message.channel.guild.id, 'makelogs') !== true || !guard.hasValue(message.author))
```
HunteRoi, line 55: Why do you add a check on the author? Wouldn't the event be fired only when an author updates it?

danny-may: Yep, I know, and yet
image
Eris has some really bad typings :( apparently there is a scenario where a message update has no author

HunteRoi (approved):

PR #349, opened 2022-06-25
Log kicks using correct moderator

danny-may:
Fixes [Modlog shows blargbot as moderator on manual kick](https://discord.com/channels/194232473931087872/987516388417605662)

### DIFF
--- a/src/cluster/managers/moderation/BanManager.ts
+++ b/src/cluster/managers/moderation/BanManager.ts
@@ -185,7 +185,8 @@ export class BanManager extends ModerationManagerBase {
         const auditLogs = await tryGetAuditLogs(member.guild, 50, undefined, AuditLogActionType.MEMBER_KICK);
         for (const log of auditLogs?.entries.values() ?? []) {
             if (log.targetID === member.id && moment(log.createdAt).isAfter(now.add(-1, 'second'))) {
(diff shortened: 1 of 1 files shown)


src/cluster/managers/moderation/BanManager.ts
```diff
@@ -185,7 +185,8 @@ export class BanManager extends ModerationManagerBase {
         const auditLogs = await tryGetAuditLogs(member.guild, 50, undefined, AuditLogActionType.MEMBER_KICK);
         for (const log of auditLogs?.entries.values() ?? []) {
             if (log.targetID === member.id && moment(log.createdAt).isAfter(now.add(-1, 'second'))) {
-                await this.modLog.logKick(member.guild, member.user, member.user, log.reason ?? undefined);
+                const moderator = log.member === undefined ? undefined : member.guild.members.get(log.member.id);
+                await this.modLog.logKick(member.guild, member.user, moderator?.user, log.reason ?? undefined);
```
HunteRoi, line 189: What happens if the moderator is undefined?

danny-may: Then it logs it as if the bot did the kick, same behaviour as all the other logs without a moderator

HunteRoi (approved):

PR #351, opened 2022-06-25
{args} and {params} debug message fix

danny-may:
Fixes [expected 1 arguments or more but got 1](https://discord.com/channels/194232473931087872/989592545417900162)

### DIFF
--- a/src/bbtag/subtags/bot/args.ts
+++ b/src/bbtag/subtags/bot/args.ts
@@ -52,7 +52,7 @@ export class ArgsSubtag extends CompiledSubtag {
             throw new NotANumberError(index);
 
         if (context.input.length <= i || i < 0)
-            throw new NotEnoughArgumentsError(i, context.input.length);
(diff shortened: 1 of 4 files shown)


src/bbtag/subtags/bot/args.ts
```diff
@@ -52,7 +52,7 @@ export class ArgsSubtag extends CompiledSubtag {
             throw new NotANumberError(index);
 
         if (context.input.length <= i || i < 0)
```
HunteRoi, line 54: Would it be possible that `i` is
1) negative?
2) how would changing the `i` value to `i+1` in the error allow the user to understand the problem?

danny-may: yep, I could be negative, theres a test for that. This is just for the debug message, garbage in garbage out imo. The error message shown for code purposes is just `Not Enough Arguments`, it makes no mention of counts

HunteRoi (approved):

PR #352, opened 2022-06-26
Derive TTL from channelid column

danny-may:
Dumps are now only generated when the embed is too long. This means the content might be null, meaning the page cant be viewed as the expiry column fails to map
image
https://blargbot.xyz/dumps/990637014720297239

### DIFF
--- a/src/database/stores/CassandraDbDumpStore.ts
+++ b/src/database/stores/CassandraDbDumpStore.ts
@@ -28,7 +28,7 @@ export class CassandraDbDumpStore implements DumpStore {
     }
 
     public async get(id: string): Promise<Dump | undefined> {
-        const dump = await this.cassandra.execute('SELECT id, content, embeds, channelid, TTL(content) as expiry FROM message_outputs WHERE id = :id', { id }, { prepare: true });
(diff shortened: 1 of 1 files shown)


src/database/stores/CassandraDbDumpStore.ts
```diff
@@ -28,7 +28,7 @@ export class CassandraDbDumpStore implements DumpStore {
     }
 
     public async get(id: string): Promise<Dump | undefined> {
-        const dump = await this.cassandra.execute('SELECT id, content, embeds, channelid, TTL(content) as expiry FROM message_outputs WHERE id = :id', { id }, { prepare: true });
+        const dump = await this.cassandra.execute('SELECT id, content, embeds, channelid, TTL(channelid) as expiry FROM message_outputs WHERE id = :id', { id }, { prepare: true });
```
HunteRoi, line 31: what is the impact on existing dumps?

danny-may: None, the TTL for all columns is the same, its just that content can be null so the TTL is null too

HunteRoi (approved):

PR #353, opened 2022-06-26
Message delete event missing some author information

danny-may:
Fixes [Message delete logs often times do not properly get the author's username and discrim](https://discord.com/channels/194232473931087872/989345216442007582) and an issue I spotted where `b!log track <sender>` didnt work

### DIFF
--- a/src/database/stores/RethinkDbGuildStore.ts
+++ b/src/database/stores/RethinkDbGuildStore.ts
@@ -725,12 +725,13 @@ export class RethinkDbGuildStore implements GuildStore {
         return true;
     }
 
-    public async setLogIgnores(guildId: string, userIds: string[]): Promise<boolean> {
+    public async setLogIgnores(guildId: string, userIds: string[], ignore: boolean): Promise<boolean> {
         const guild = await this.#table.get(guildId);
         if (guild === undefined)
             return false;
 
-        if (!await this.#table.update(guildId, r => ({ logIgnore: r('logIgnore').default([]).setUnion(userIds) })))
+        const method = ignore ? 'setUnion' : 'setDifference';
+        if (!await this.#table.update(guildId, r => ({ logIgnore: r('logIgnore').default([])[method](userIds) })))
             return false;
 
         setProp(guild, 'logIgnore', [...new Set([...guild.logIgnore ?? [], ...userIds])]);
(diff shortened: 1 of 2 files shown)


src/database/stores/RethinkDbGuildStore.ts
```diff
@@ -725,12 +725,13 @@ export class RethinkDbGuildStore implements GuildStore {
...
 
-    public async setLogIgnores(guildId: string, userIds: string[]): Promise<boolean> {
+    public async setLogIgnores(guildId: string, userIds: string[], ignore: boolean): Promise<boolean> {
         const guild = await this.#table.get(guildId);
         if (guild === undefined)
             return false;
 
-        if (!await this.#table.update(guildId, r => ({ logIgnore: r('logIgnore').default([]).setUnion(userIds) })))
+        const method = ignore ? 'setUnion' : 'setDifference';
+        if (!await this.#table.update(guildId, r => ({ logIgnore: r('logIgnore').default([])[method](userIds) })))
```
HunteRoi, line 734: any possibility this line can be made easier to understand? 😳

src/database/stores/RethinkDbGuildStore.ts
```diff
@@ -725,12 +725,13 @@ export class RethinkDbGuildStore implements GuildStore {
         return true;
     }
 
-    public async setLogIgnores(guildId: string, userIds: string[]): Promise<boolean> {
+    public async setLogIgnores(guildId: string, userIds: string[], ignore: boolean): Promise<boolean> {
```
HunteRoi, line 728: you updated the signature of this method but I don't see any change in the `GuildStore` interface nor in the `log` command, do I? Why is that?

danny-may: The interface always had the `ignore:boolean` param, the impl just wasnt using it. whoops!

src/database/stores/RethinkDbGuildStore.ts
danny-may: I could split it onto multiple lines? But its rethink, its not the most elegant of syntaxes in general

danny-may: Added comments

HunteRoi (approved):

PR #354, opened 2022-06-26
Fix shrinkwrapping

danny-may:
Fixes [shrinkwrap install doesnt seem to work](https://discord.com/channels/194232473931087872/989337953581015050)


HunteRoi (approved):

HunteRoi:
Just a remark: wouldn't it be easier to deconstruct the command object into the shrinkwrap object, instead of just setting properties on the fly?

danny-may:
There are some properties on the command that we dont want to put into the shrinkwrap, like the author and authorizer. doing it this way ensures excess properties arent added. spreads are fine for stuff that is remaining internal, but we should be explicit imo if we are sending stuff to users, just to be safe

PR #364, opened 2022-07-15
Add documentation manager, using interactions

### LINKED ISSUE #336: Suggestion - b!docs UX improvements
I have a few suggestions regarding the UX of b!docs.

- Under b!docs, add buttons or a select menu allowing the user to switch to, e.g. b!docs subtags instantly without having to run the command again.
- Allow users to continue switching between categories after they've selected one. Currently, to switch to another category, you have to run b!docs again, which is inconvenient, especially when you're working on something.
- To add to that, perhaps save the previous category selected and display that when b!docs subtags is run again.
- Add a search function. Possible implementation would be a button that pops up a modal which prompts the user for a search term when clicked. This will help users search through the categories for a subtag that they need.
- Perhaps allow multiple selections within the selector (which Discord supports) to view multiple categories at once, which would be extremely useful when working on a large project. This could perhaps be saved and shown again when b!docs subtags is run again.

danny-may:
Replaces help and bbtag docs with a new framework making use of interactions to make the documentation shorter and navigatable.

e.g.
image

EDIT by @HunteRoi :
Fixes #336


src/cluster/dcommands/general/tag.ts
```diff
@@ -163,6 +167,9 @@ export class TagCommand extends GuildCommand {
                 }
             ]
         });
+
+        this.#docs = new BBTagDocumentationManager(cluster, 'tag');
```
HunteRoi, line 171: why are you not providing the `t` alias here?

src/cluster/managers/documentation/BBTagDocumentationManager.ts
```diff
@@ -0,0 +1,336 @@
+import { bbtag, limits, Subtag, SubtagSignature, SubtagSignatureValueParameter, SubtagType, tagTypeDetails, tagVariableScopeProviders } from '@blargbot/bbtag';
```
HunteRoi, line 1: General comment about this:
> isn't the information duplicated between the documentation manager and the website by the way?

src/cluster/managers/documentation/CommandDocumentationManager.ts
```diff
@@ -0,0 +1,225 @@
...
+            case 'custom': return 0x7289da;
+            case 'general': return 0xefff00;
+            case 'nsfw': return 0x010101;
+            case 'image': return 0xefff00;
+            case 'admin': return 0xff0000;
+            case 'social': return 0xefff00;
+            case 'owner': return 0xff0000;
+            case 'developer': return 0xff0000;
+            case 'staff': return 0xff0000;
+            case 'support': return 0xff0000;
```
HunteRoi, line 174: I think these colors are already duplicated in an enum or something. Might be worth taking a look at other files, right?

src/core/utils/iter/index.ts
```diff
@@ -0,0 +1,5 @@
+import * as expand from './expand';
+
+export const iter = {
+    ...expand
+};
```
HunteRoi, line 5: what's the goal here?

src/cluster/managers/documentation/DocumentationTreeManager.ts
```diff
@@ -0,0 +1,89 @@
...
+                name: `${d.name} - ${i.name}`,
+                embed: {
+                    ...i.embed,
+                    color: i.embed.color ?? d.embed.color
+                }
+            }));
+        });
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
HunteRoi, line 26: I guess that is because you are implementing a method from the parent with a custom logic and you don't need th user and channel parameters, right?

src/cluster/dcommands/general/tag.ts
danny-may: this `tag` is used to build the custom_id for the buttons/selects and is used to identify what should handle the interaction. Cant have multiple of those, and I wanted them to be ~3 chars. Will be using a similar system for handling interactions going forwards too I believe

src/cluster/managers/documentation/BBTagDocumentationManager.ts
danny-may: Yep, but only the copy. The documentation for the commands/subtags themselves are all generated from the same data on both the website and here

src/core/utils/iter/index.ts
danny-may: I forgot to remove this, whoops

src/cluster/managers/documentation/CommandDocumentationManager.ts
danny-may: They are in a few places. I think doing a color consolidtation pass would be best left to another PR though

src/cluster/managers/documentation/DocumentationTreeManager.ts
danny-may: It was origionally going to be a method that was meant to be overridden by a derived type, but then I made it a private method and forgot to remove the params. Fixed now

HunteRoi (approved):

PR #366, opened 2022-07-16
Centralize guild loading & remove hello message 2.0

danny-may:
Hopefully fixes an OOM issue introduced by #361

### DIFF
--- a/src/cluster/managers/GuildManager.ts
+++ b/src/cluster/managers/GuildManager.ts
@@ -1,0 +1,65 @@
+import { metrics } from '@blargbot/core/Metrics';
+import { Guild } from 'eris';
+import moment from 'moment-timezone';
+
+import { Cluster } from '../Cluster';
+
+export class GuildManager {
+    #blacklist: Record<string, boolean | undefined> | undefined;
+
+    public constructor(
+        private readonly cluster: Cluster
+    ) {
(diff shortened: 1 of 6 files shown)


src/cluster/managers/GuildManager.ts
```diff
@@ -0,0 +1,65 @@
...
+        const blacklist = await this.#getBlacklist();
+        if (blacklist[guild.id] === true) {
+            await this.#guildBlacklisted(guild);
+            return;
+        }
+
+        if (await this.cluster.database.guilds.upsert(guild) === 'inserted') {
+            await this.cluster.util.ensureMemberCache(guild);
+            const users = guild.members.filter(m => !m.user.bot).length;
+            const bots = guild.members.filter(m => m.user.bot).length;
```
HunteRoi, line 52: wouldn't it be faster to have a single filter and a math operation?

```suggestion
            const users = guild.members.filter(m => !m.user.bot).length;
            const bots = guild.members.size - users;
```

HunteRoi:
Also not sure about the veracity of the name of this PR ^^

src/cluster/managers/GuildManager.ts
danny-may: That sounds too big brain for me

HunteRoi (approved):

RagingLink (approved):

PR #371, opened 2022-07-31
Filter should trim arg before parsing

danny-may:
Fixes [broken {filter}](https://discord.com/channels/194232473931087872/1003000055663427704)

### DIFF
--- a/src/bbtag/subtags/array/filter.ts
+++ b/src/bbtag/subtags/array/filter.ts
@@ -31,7 +31,7 @@ export class FilterSubtag extends CompiledSubtag {
             for (const item of array) {
                 await context.limit.check(context, 'filter:loops');
                 await context.variables.set(varName, item);
-                if (parse.boolean(await code.execute()) === true)
+                if (parse.boolean((await code.execute()).trim()) === true)
                     yield item;
 
                 if (context.data.state !== BBTagRuntimeState.RUNNING)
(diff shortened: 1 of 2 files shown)


src/bbtag/subtags/array/filter.ts
```diff
@@ -31,7 +31,7 @@ export class FilterSubtag extends CompiledSubtag {
             for (const item of array) {
                 await context.limit.check(context, 'filter:loops');
                 await context.variables.set(varName, item);
-                if (parse.boolean(await code.execute()) === true)
+                if (parse.boolean((await code.execute()).trim()) === true)
```
HunteRoi, line 34: Can you elaborate on why trimming the execution's result actually fixes the problem of having a `{//;comment}` within the filter ?

danny-may: When evaluating args, leading and trailing spaces are cut off. Adding a comment prevents this because the space after the comment doesnt get trimmed. Comments in bbtag arent true comments, they do change the behaviour of bbtag because they are also subtags, they just dont do anything with their args.
image

If you have a comment inside your condition with whitespace between it and your result, that whitespace will be preserved. It needs to be trimmed before parsing. This is also what the js codebase did, so this is a fix for a breaking change from the update

HunteRoi (approved):

PR #372, opened 2022-07-31
Implement poll announcements

danny-may:
Fixes [poll command does not work when '-a' or '--announce' flag used](https://discord.com/channels/194232473931087872/989132399155748884)

### DIFF
--- a/src/cluster/dcommands/admin/announce.ts
+++ b/src/cluster/dcommands/admin/announce.ts
@@ -1,8 +1,8 @@
 import { GuildCommand } from '@blargbot/cluster/command';
 import { GuildCommandContext } from '@blargbot/cluster/types';
 import { CommandType, discord } from '@blargbot/cluster/utils';
-import { guard, humanize } from '@blargbot/core/utils';
-import { AllowedMentions, Constants, EmbedOptions, KnownChannel, KnownGuildTextableChannel, Role } from 'eris';
+import { humanize } from '@blargbot/core/utils';
+import { AllowedMentions, Constants, EmbedOptions, KnownChannel, Role } from 'eris';
 import moment from 'moment-timezone';
 
 export class AnnounceCommand extends GuildCommand {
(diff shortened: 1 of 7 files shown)


src/cluster/managers/AnnouncementManager.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+type AnnounceResult =
+    | Result<'Success', AnnouncementConfig>
+    | Result<'NotAllowed'>
+    | Result<'ChannelNotFound'>
+    | Result<'ChannelNotInGuild'>
+    | Result<'ChannelInvalid'>
+    | Result<'RoleNotFound'>
+    | Result<'TimedOut'>
```
HunteRoi, line 108: While I understand those are only used here, I am not sure we follow the same conventions regarding other managers. How can we refactor our code to fit a standardized way of making managers?

src/cluster/dcommands/admin/announce.ts
```diff
@@ -98,73 +99,13 @@ export class AnnounceCommand extends GuildCommand {
...
-    }
-
-    private async configureCore(context: GuildCommandContext, channel: KnownChannel | undefined, role: Role | undefined): Promise<boolean | string> {
-        if (channel === undefined) {
-            const result = await context.queryChannel({
-                choices: context.channel.guild.channels.filter(guard.isTextableChannel).values(),
-                prompt: this.info('Please select the channel that announcements should be put in.')
-            });
+        const channelStr = config.channel?.mention ?? '`<unconfigured>`';
+        const roleStr = config.role?.mention ?? '`<unconfigured>`';
```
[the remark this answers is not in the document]
danny-may: Yep, the check above checks if both are undefined. If only 1 is undefined then we need to handle that too

src/cluster/managers/AnnouncementManager.ts
danny-may: I think having it like this is better, the types should be placed where they are used. The `types.ts` files arent the best way to do things imo, but when the entire solution is being rewritten, it was definitely the easiest. Imo the only time types should be placed in a `types.ts` going forwards is when they are an abstract concept, like the `Result` type being used here.

HunteRoi (approved):

PR #373, opened 2022-08-20
Fix issue with smartSplit and unpaired quotes

danny-may:
Fixes [{json} behaving strangely inside {inject}](https://discord.com/channels/194232473931087872/1007124647923425371)

### DIFF
--- a/src/core/utils/humanize/smartSplit.ts
+++ b/src/core/utils/humanize/smartSplit.ts
@@ -24,7 +24,7 @@ export function* smartSplitRanges(source: string): Generator<{ start: number; en
 }
 
 interface SmartSplitItem {
-    readonly ranges: CharRange[];
+    readonly ranges: readonly CharRange[];
     readonly content: string;
     readonly start: number;
     readonly end: number;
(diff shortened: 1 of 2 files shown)


src/core/utils/humanize/smartSplit.ts
```diff
@@ -47,70 +69,123 @@ function* smartSplitIterLimit(source: string, limit: number): Generator<string>
...
-            case ' ': {
-                if (inQuote)
-                    builder.push(i);
-                else if (start !== undefined) {
-                    yield createSplitItem(source, builder, start, i);
-                    start = undefined;
-                    builder = [];
-                }
+            case '\\':
+                yield* yieldBlock();
```
HunteRoi, line 94: It seems like this could be exported outside the switch, right?

HunteRoi (changes requested):

src/core/utils/humanize/smartSplit.ts
```diff
@@ -47,70 +69,123 @@ function* smartSplitIterLimit(source: string, limit: number): Generator<string>
...
-        }));
-
-    return {
-        ranges: ranges,
-        content: ranges.map(r => r.content).join(''),
-        end,
-        start
     };
+
+    for (; ctx.index < ctx.tokenCount; ctx.index++)
```
HunteRoi, line 135: This loop feels like a hack🤭 Wouldn't be better to just use a while?

src/core/utils/humanize/smartSplit.ts
danny-may: We dont always want to do it. The default case doesnt call this method. I didnt want to add in another check, so pulling it out to a single method was the best option I felt

HunteRoi (changes requested):

src/core/utils/humanize/smartSplit.ts
danny-may: I did have it as a while at first, but it is more akin to a for loop. you have a counter which increments each loop, and a limit on that counter. The only reason we dont have a `var i = 0` bit is because the loop index can be changed by the method it calls so the value needs to be stored on a reference type

HunteRoi (approved):

PR #374, opened 2022-08-20
User doesnt have to be in the guild to ban them

danny-may:
Implements the change requested here: [b!ban doesnt work with userids if the user isnt in the guild](https://discord.com/channels/194232473931087872/1009354648928403467)


HunteRoi:
I thought this was normal and people had to use hackban ?

danny-may:
Yeah, I guess this is more of a change than a bug fix. The only difference between the `user` binding and the `member` binding is that the `user` binding can accept an id of someone not on the guild. Banning by username/nickname still needs the user to be on the same guild as the lookup system is restricted by guild members

HunteRoi (approved):

PR #378, opened 2022-08-27
Use private fields not private modifier

danny-may:
Just a bit of cleanup. The private keyword in TS doesnt actually make the field private at runtime. The #fields enforce this


danny-may:
Converting to a draft as this PR got a bit messy with other stuff. Ill clean it up and split it into 2 PRs

src/cluster/managers/awaiters/Awaiter.ts
```diff
@@ -1,54 +1,60 @@
...
             (pools[poolId] ??= []).push(this);
 
-        this.timeout = setTimeout(() => this.cancel(), timeout);
+        this.#timeout = setTimeout(() => this.cancel(), timeout);
     }
 
     public async wait(): Promise<T | undefined> {
-        return await this.pcs.promise;
+        return await this.#pcs.promise;
     }
```
HunteRoi, line 29: not related to the private fields change, but wouldn't this be changed to something like:
```ts
    public wait(): Promise<T | undefined> {
        return this.#pcs.promise;
    }
```

HunteRoi (changes requested):

src/cluster/managers/awaiters/Awaiter.ts
danny-may: it could, but in general im not a fan of returning promises. It hides methods from the call stack in case of an exception.

HunteRoi (approved):

PR #379, opened 2022-08-27
channeledit/create json should allow strings

danny-may:
Fixes [{channeledit} seems broken... returns 'Invalid JSON' when trying to set rateLimitPerUser](https://discord.com/channels/194232473931087872/1012176276385185853)


HunteRoi (approved):
I still think the mapping things should be extracted to their own library. We already have them extracted to their own subrepository. Is there something preventing you @Titansmasher to move them to a dedicated repository of yours?

danny-may:
Nope, nothing stopping me except laziness and coming up with a good name 😄 I might do it this weekend

PR #380, opened 2022-08-28
Potentially fix member edit race condition

danny-may:
It would seem that calling `Member#edit` doesnt cause eris to ensure the member is edited before resolving the promise. I think this is causing a race condition in some situations. This PR hopefully fixes [Something screwy going on with {jget} or {roleadd} and {roleremove}](https://discord.com/channels/194232473931087872/1012147398333706363) but will need testing once deployed to be sure

### DIFF
--- a/src/bbtag/subtags/role/roleadd.ts
+++ b/src/bbtag/subtags/role/roleadd.ts
@@ -64,7 +64,9 @@ export class RoleAddSubtag extends CompiledSubtag {
             return false;
 
         try {
-            await member.edit({ roles: member.roles.concat(...roles.map(r => r.id)) }, context.auditReason());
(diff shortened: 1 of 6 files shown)


HunteRoi:
Shouldn't this be taken up to https://github.com/abalabahaha/eris ?

HunteRoi (approved):

danny-may:
Maybe, they dont do this anywhere in the code base, so maybe not. In my opinion the real fix is needed here:
https://github.com/abalabahaha/eris/blob/dev/lib/Client.js#L1570
This shouldnt be a new member, it should edit the already cached member if one exists. But it will be forever before that change gets done, so doing this here fixes it for us for now

PR #382, opened 2022-08-30
Allow logs, tidy and tag should use queryUser

danny-may:
Logs, tidy and tag dont rely on a user being on the guild, so they should use queryUser.


danny-may:
Please excuse the additional files that got included. Github actions are using a newer version of node than they used to which doesnt work with the current version of yarn berry. I had to update yarn and typescript, which came with a few compile errors.

src/cluster/command/CommandContext.ts
```diff
@@ -103,9 +103,20 @@ export class CommandContext<TChannel extends KnownTextableChannel = KnownTextabl
...
     }
+    public async queryUser(options: SlimEntityFindQueryOptions): Promise<ChoiceQueryResult<User>>;
+    public async queryUser(this: GuildCommandContext, options: Omit<SlimEntityFindQueryOptions, 'guild'>): Promise<ChoiceQueryResult<User>>;
+    public async queryUser(options: SlimEntityPickQueryOptions<User>): Promise<ChoiceQueryResult<User>>;
+    public async queryUser(options: SlimEntityQueryOptions<User> | Omit<SlimEntityFindQueryOptions, 'guild'>): Promise<ChoiceQueryResult<User>> {
+        if ('choices' in options)
+            return await this.util.queryUser({ ...options, context: this.message, actors: this.author });
+
+        if ('guild' in options)
+            return await this.util.queryUser({ ...options, context: this.message, actors: this.author });
```
HunteRoi, line 114: could be merged into a single condition, right?

src/cluster/command/CommandContext.ts
```diff
@@ -103,9 +103,20 @@ export class CommandContext<TChannel extends KnownTextableChannel = KnownTextabl
 
         throw new Error('Cannot queryMember without a guild!');
     }
+    public async queryUser(options: SlimEntityFindQueryOptions): Promise<ChoiceQueryResult<User>>;
+    public async queryUser(this: GuildCommandContext, options: Omit<SlimEntityFindQueryOptions, 'guild'>): Promise<ChoiceQueryResult<User>>;
```
HunteRoi, line 107: what's the point in overwriting the this here?

HunteRoi (changes requested):

src/cluster/command/CommandContext.ts
danny-may: sadly not. Logically yes, but typescript complains if you do
image

src/cluster/command/CommandContext.ts
danny-may: Its to add the overloads that are accepted. The second overload is for the case where there is a guild associated with the context, so you dont need to supply one.

HunteRoi (approved):

PR #384, opened 2022-09-01
Add extension parameter to the bbtag raw commands

danny-may:
Allows you to specify the file type for the `raw` commands when getting bbtag.
Implements [#814](https://airtable.com/appKTYe5luRw2aoas/tblyFuWE6fEAbaOfo/viwDg5WovcwMA9NIL/reczTF7VBRsv9mmAQ)

### DIFF
--- a/src/cluster/dcommands/admin/autoresponse.ts
+++ b/src/cluster/dcommands/admin/autoresponse.ts
@@ -125,7 +125,7 @@ export class AutoResponseCommand extends GuildCommand {
         const responseBase = `The raw code for ${match.id === 'everything' ? 'the everything autoresponse' : `autoresponse ${match.id}`} is`;
 
         const response = this.success(`${responseBase}:\n${codeBlock(match.ar.content)}`);
-        return guard.checkMessageSize(response)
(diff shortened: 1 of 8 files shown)


src/cluster/dcommands/admin/autoresponse.ts
```diff
@@ -125,13 +125,13 @@ export class AutoResponseCommand extends GuildCommand {
         const responseBase = `The raw code for ${match.id === 'everything' ? 'the everything autoresponse' : `autoresponse ${match.id}`} is`;
 
         const response = this.success(`${responseBase}:\n${codeBlock(match.ar.content)}`);
-        return guard.checkMessageSize(response)
+        return !match.ar.content.includes('`') && guard.checkMessageSize(response)
```
HunteRoi, line 128: why do you check for no backticks in the content?

danny-may: Because backticks inside a codeblock might break the codeblock, and if we add escapes then the codeblock technically no longer shows the raw of the tag. Easier and less error prone to just send the file in that case

HunteRoi (approved):

PR #390, opened 2022-09-04
Fix smartsplit with escapes inside quote

danny-may:
Fixes [Json funkiness in {exec}](https://discord.com/channels/194232473931087872/1015443632834220032)

### DIFF
--- a/src/core/utils/humanize/smartSplit.ts
+++ b/src/core/utils/humanize/smartSplit.ts
@@ -37,7 +37,6 @@ interface CharRange {
 }
 
 interface SmartSplitTokenContext {
-    readonly source: string;
     readonly ranges: CharRange[];
     index: number;
     readonly tokenCount: number;
@@ -48,14 +47,14 @@ interface SmartSplitTokenContext {
 const enum SmartSplitTokenType {
     LITERAL,
     QUOTE,
-    BREAK,
-    ESCAPE
+    BREAK
 }
 
 interface SmartSplitToken {
     readonly type: SmartSplitTokenType;
     readonly start: number;
     readonly end: number;
+    get content(): string;
 }
 
 function* smartSplitIterLimit(source: string, limit: number): Generator<string> {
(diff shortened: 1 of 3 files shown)


src/core/utils/humanize/smartSplit.ts
```diff
@@ -84,15 +90,22 @@ function* tokenize(source: string): Generator<SmartSplitToken> {
...
-                yield { type: SmartSplitTokenType.QUOTE, start: i, end: i + 1 };
+                yield { type: SmartSplitTokenType.QUOTE, start: i, end: i + 1, content: '"' };
                 break;
             case '\\':
                 yield* yieldBlock();
-                yield { type: SmartSplitTokenType.ESCAPE, start: i, end: i + 1 };
+                yield {
+                    type: SmartSplitTokenType.LITERAL,
+                    start: i,
+                    end: ++i + 1,
```
HunteRoi, line 104: why do you double increment here?

src/core/utils/humanize/smartSplit.ts
```diff
@@ -84,15 +90,22 @@ function* tokenize(source: string): Generator<SmartSplitToken> {
...
+                    get content() {
+                        return source.slice(this.start + 1, this.end);
```
HunteRoi, line 106: why do you increment here too?

src/core/utils/humanize/smartSplit.ts
danny-may: This token is 2 chars wide, the backslash and the following character.
Because this "consumes" the next character too i needs to be incremented so the next iteration doesnt consider it

src/core/utils/humanize/smartSplit.ts
danny-may: The backslash of an escape doesnt form part of its literal output, so we want to slice with an offset of 1 to exclude it

HunteRoi (approved):

PR #401, opened 2022-09-15
Add limits on command names and handle large command counts

danny-may:
Fixes [b!help sends nothing when there are too many commands](https://discord.com/channels/194232473931087872/1019673077086425151)

I added a character limit on command names to prevent them from breaking the help documentation. The names get put into the label field of the dropdowns, which discord limits to 100 characters.

The migration changes are designed to be expandable in the future, and to only ever be run once. I might restructure it later to be reusable for the other tables, but at the moment only the guild table needs it.
Theres no concerns about race conditions with the migrations as a cluster/master isnt classed as ready until the database is connected which includes the migrations having been run. To be safe, we should `restart kill` to deploy this codebase however.

### DIFF
--- a/src/cluster/dcommands/admin/ccommand.ts
+++ b/src/cluster/dcommands/admin/ccommand.ts
@@ -752,10 +752,13 @@ export class CustomCommandCommand extends GuildCommand {
             return undefined;
 
         const command = await context.database.guilds.getCommand(context.channel.guild.id, commandName);
-        if (command === undefined)
-            return { name: commandName };
+        if (command !== undefined)
+            return { name: command.name, command };
 
-        return { name: command.name, command };
+        if (commandName.length > 80)
+            return this.error('Command names cannot be longer than 80 characters');
+
+        return { name: commandName };
     }
 }
 
--- a/src/cluster/dcommands/admin/ccommand.ts
+++ b/src/cluster/dcommands/admin/ccommand.ts
@@ -755,8 +755,8 @@ export class CustomCommandCommand extends GuildCommand {
         if (command !== undefined)
             return { name: command.name, command };
 
-        if (commandName.length > 80)
-            return this.error('Command names cannot be longer than 70 characters');
+        if (commandName.length > 100)
+            return this.error('Command names cannot be longer than 100 characters');
 
         return { name: commandName };
     }
--- a/src/cluster/dcommands/admin/ccommand.ts
+++ b/src/cluster/dcommands/admin/ccommand.ts
@@ -2,7 +2,7 @@ import { bbtag } from '@blargbot/bbtag';
 import { Cluster } from '@blargbot/cluster';
 import { GuildCommand } from '@blargbot/cluster/command';
 import { CommandResult, CustomCommandShrinkwrap, GuildCommandContext, GuildShrinkwrap, ICommand, SignedGuildShrinkwrap } from '@blargbot/cluster/types';
-import { codeBlock, CommandType, guard, humanize, parse } from '@blargbot/cluster/utils';
+import { codeBlock, CommandType, guard, humanize, parse, snowflake } from '@blargbot/cluster/utils';
 import { Configuration } from '@blargbot/config';
 import { SendContent, SendPayload } from '@blargbot/core/types';
 import { FlagDefinition, NamedGuildCommandTag, NamedGuildSourceCommandTag } from '@blargbot/domain/models';
@@ -226,7 +226,7 @@ export class CustomCommandCommand extends GuildCommand {
         if (typeof match !== 'object')
             return match;
 
-        return await this.#saveCommand(context, 'created', match.name, content);
+        return await this.#saveCommand(context, 'created', undefined, match.name, content);
     }
 
     public async editCommand(context: GuildCommandContext, commandName: string | undefined, content: string | undefined): Promise<string | undefined> {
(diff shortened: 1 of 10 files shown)


src/cluster/dcommands/admin/ccommand.ts
```diff
@@ -752,10 +756,13 @@ export class CustomCommandCommand extends GuildCommand {
...
 
         const command = await context.database.guilds.getCommand(context.channel.guild.id, commandName);
-        if (command === undefined)
-            return { name: commandName };
+        if (command !== undefined)
+            return { name: command.name, command };
 
-        return { name: command.name, command };
+        if (commandName.length > 100)
+            return this.error('Command names cannot be longer than 100 characters');
```
HunteRoi, line 763: I feel like this is not necessary in this part of the code. Might be useful higher in the logic though

src/cluster/managers/documentation/CommandDocumentationManager.ts
```diff
@@ -88,6 +88,31 @@ export class CommandDocumentationManager extends DocumentationTreeManager {
         };
     }
 
+    #listCommandNames(names: readonly string[]): string {
+        const charLimit = discord.getLimit('embed.field.value') - 6;
```
HunteRoi, line 92: `- 6` ? magic number aaaaaaaaaaaaaaaaa

src/cluster/managers/documentation/CommandDocumentationManager.ts
```diff
@@ -88,6 +88,31 @@ export class CommandDocumentationManager extends DocumentationTreeManager {
...
+
+        const remaining = [...names];
+        let missing = 0;
+        const postfix1 = '+ ';
+        const postfix2 = ' more';
+        function contentLength(): number {
+            return remaining.reduce((p, c) => p + c.length + 2, 0) - 2
+                + postfix1.length
+                + postfix2.length
+                + missing.toString().length;
```
HunteRoi, line 104: not a big thing but I would have placed the missing line before the postfix2 even though it does not change the value returned by contentLength()

src/cluster/managers/documentation/CommandDocumentationManager.ts
```diff
@@ -88,6 +88,31 @@ export class CommandDocumentationManager extends DocumentationTreeManager {
         };
     }
 
+    #listCommandNames(names: readonly string[]): string {
+        const charLimit = discord.getLimit('embed.field.value') - 6;
+        if (names.reduce((p, c) => p + c.length + 2, 0) - 2 <= charLimit)
```
HunteRoi, line 93: why you adding 2 in the reduce then removing 2 from the total?

src/cluster/managers/documentation/CommandDocumentationManager.ts
```diff
@@ -88,6 +88,31 @@ export class CommandDocumentationManager extends DocumentationTreeManager {
...
+                + postfix1.length
+                + postfix2.length
+                + missing.toString().length;
+        }
+        while (contentLength() > charLimit) {
+            remaining.pop();
+            missing++;
+        }
+        if (remaining.length === 0)
+            return `${missing} ${pluralise(missing, 'command')}`;
```
HunteRoi, line 111: So if the whole content is still too large, you just send (for example) `15 commands` as the returned value?

src/database/stores/RethinkDbGuildStore.ts
```diff
@@ -788,13 +791,59 @@ export class RethinkDbGuildStore implements GuildStore {
...
+                    if (ccommand?.id === undefined) {
+                        ccommands[name] = { id: snowflake.create().toString() };
+                    }
+                }
+                await table.update(guild.guildid, { ccommands });
+            }
+        }
+    }
+];
+
```
HunteRoi, line 846: wouldn't it be better to structure this? we might need it more than once (actually it's the second time we run a migration so a third is probably not avoidable in the future

HunteRoi (changes requested):

src/cluster/managers/documentation/CommandDocumentationManager.ts
danny-may: 6 comes from the padding of 3 backticks added to each side of the string, guess I could use a .length instead though

src/cluster/managers/documentation/CommandDocumentationManager.ts
danny-may: Thats what happens to the length when you .join and array, the filler gets added between each element, so it gets added `.length - 1` times
`['a','b','c'].join(', ') === 'a, b, c'`
`('a'.length + 2) + ('b'.length + 2) + ('c'.length + 2) - 2`

src/cluster/managers/documentation/CommandDocumentationManager.ts
danny-may: Fair point!

src/cluster/managers/documentation/CommandDocumentationManager.ts
danny-may: Correct, this is for the case where the first element is > 1024 characters long, and so we cannot show any inside the code block

src/cluster/managers/documentation/CommandDocumentationManager.ts
danny-may: Ive refactored this entire method to be more readable, thoughts?

src/database/stores/RethinkDbGuildStore.ts
danny-may: I think ill work on a proper system for doing migrations after this, I think doing it in this PR would be a bit of scope creep. This was the bare minimum needed for this to become functional, so for now ill leave it as this.

src/cluster/dcommands/admin/ccommand.ts
danny-may: I think it makes sense here, it fits with the pattern of validation for the other bits, and going forwards this 100 character limit should be applied to any new ccommands that get made, for which every route goes through here.
I would have put it at the start of this method, but that would then prevent you interacting with any commands that already exist over 100 characters, so doing the check only when the command doesnt already exist is the best place for it imo.

PR #408, opened 2022-09-29
Update package versions

danny-may:


danny-may:
Seems typescript 4.8.4 isnt supported by yarn berry yet, waiting until their fix is deployed before this PR can go through
https://github.com/yarnpkg/berry/issues/4915

HunteRoi:
I saw you updated mocha to 10.0.0. Isn't there breaking changes ? I don't see any update in test files... 🤔

danny-may:
Mocha was updated in the last dependency update I did, this is just the update to the typings

HunteRoi:
@danny-may What about the builds failing?

HunteRoi (approved):

danny-may:
Thats due to the yarn issue. Basically because dependencies are zipped files in `.yarn/cache` rather than normal folders in `node_modules`, yarn needs to inject in the logic to typescript to enable module lookups. 
Without the fix from yarn, typescript cannot find any of the projects dependencies and so fails to build

PR #422, opened 2022-11-01
Add coalesce operator

danny-may:
Adds subtag operator requested here: [Null coalesce operator](https://discord.com/channels/194232473931087872/1036967327264157716)

### DIFF
--- a/src/bbtag/compilation/compileSignatures.ts
+++ b/src/bbtag/compilation/compileSignatures.ts
@@ -12,7 +12,10 @@ export function compileSignatures(signatures: readonly SubtagSignatureCallable[]
     let min = initialResolver;
     let max = initialResolver;
 
-    for (const signature of signatures) {
+    // Named signatures first as they are more restrictive
+    const orderedSignatures = [...signatures].sort((a, b) => a.subtagName === b.subtagName ? 0 : a.subtagName === undefined ? 1 : -1);
+
+    for (const signature of orderedSignatures) {
         if (!('implementation' in signature))
             continue;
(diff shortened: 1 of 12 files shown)


src/bbtag/compilation/compileSignatures.ts
```diff
@@ -43,10 +46,12 @@ export function compileSignatures(signatures: readonly SubtagSignatureCallable[]
...
 function createConditionalHandler(signature: SubtagSignatureCallable, resolver: ArgumentResolver): ConditionalSubtagHandler {
+    const name = signature.subtagName?.toLowerCase();
+
     return {
-        canHandle(subtag) {
-            return resolver.isExactMatch(subtag);
-        },
+        canHandle: name === undefined
+            ? subtag => resolver.isExactMatch(subtag)
+            : (subtag, subtagName) => subtagName.toLowerCase() === name && resolver.isExactMatch(subtag),
```
HunteRoi, line 54: what's the point in a ternary operation here? you could merge the conditions together and only have it return the boolean result from everything with some || and &&

HunteRoi (changes requested):

src/bbtag/compilation/compileSignatures.ts
danny-may: The idea is that the body of this function is a hot path, every subtag call will call whatever this function is. The name never changes and we dont care what the subtag name is if the signatures subtag name is undefined, so why check it every time?

Its a bit of a micro optimisation for sure, but we want this as fast as possible

HunteRoi (approved):

PR #426, opened 2022-11-03
Sync from crowdin rather than github

danny-may:
Pull translations from crowdin on startup rather than having them checked in to github. They can be reloaded on demand with `b!i18n reload`

NOTE: Dont merge this PR, some changes in crowdin need to happen in parallel when it does get merged


HunteRoi (approved):

src/core/BaseUtilities.ts
```diff
@@ -23,17 +22,12 @@ export class BaseUtilities {
...
     public constructor(
         public readonly client: BaseClient
     ) {
-        this.translator = new FileSystemTranslationSource(
-            path.join(
-                path.dirname(require.resolve('@blargbot/res/package')),
-                'i18n'
-            )
-        );
+        this.translator = new CrowdinTranslationSource('a713aad8fe135bf923f9587yoka');
```
HunteRoi, line 30: what's that key there?

danny-may: Thats the public distribution key, used to say what bundle of files from the CDN we want

danny-may: I could put it in the config, but theres no need to keep it secret and I doubt this will ever change

PR #430, opened 2022-11-03
Use file not files

danny-may:
When sending files, blargbot strips off the `files` property because that is sent in a second parameter in eris. This is different for editing messages, where eris checks for a `file` property on the message payload. Dont ask why, its a bit nuts. This brings all of blargbot in line with eris, using the `file` property rather than `files`

Typing in eris:
image
image


HunteRoi:
thanks Eris

HunteRoi (approved):

PR #431, opened 2022-11-04
Add locale name from language definition

danny-may:
Language names should be in the language its describing rather than english

### DIFF
--- a/src/crowdin/index.ts
+++ b/src/crowdin/index.ts
@@ -67,7 +67,11 @@ export class CrowdinTranslationSource implements ITranslationSource {
             this.#languageKeys.delete(language.locale);
             this.#languageData.delete(language.locale);
         } else {
-            this.#languageKeys.set(language.locale, { name: language.name, keys: new Set(lookup.keys()) });
+            this.#languageKeys.set(language.locale, {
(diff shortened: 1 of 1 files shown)


src/crowdin/index.ts
```diff
@@ -67,7 +67,11 @@ export class CrowdinTranslationSource implements ITranslationSource {
             this.#languageKeys.delete(language.locale);
             this.#languageData.delete(language.locale);
         } else {
-            this.#languageKeys.set(language.locale, { name: language.name, keys: new Set(lookup.keys()) });
+            this.#languageKeys.set(language.locale, {
+                name: lookup.get('config.localeName') ?? language.name,
+                keys: new Set(lookup.keys())
+            });
+            lookup.delete('config.localeName');
```
HunteRoi, line 74: why do you delete it?

danny-may: Just so it doesnt contribute to any stats, this is the only place its used

HunteRoi: what's the goal of the lookup object actually?

danny-may: It contains the mapping of keys to translated templates, so the english templates can be swapped out

HunteRoi (approved):

PR #434, opened 2022-11-20
Add support for paged bans

danny-may:
Fixes [{guildbans} still only shows the first 1,000 bans](https://discord.com/channels/194232473931087872/1040950212870549635)

### DIFF
--- a/src/core/BaseUtilities.ts
+++ b/src/core/BaseUtilities.ts
@@ -633,8 +633,9 @@ export class BaseUtilities {
     }
 
     async #ensureGuildBans(guild: Guild): Promise<void> {
-        for await (const _ of this.streamAllBans(guild))
-            ;
+        for await (const _ of this.streamAllBans(guild)) {
+            // NO-OP - streamAllBans caches each record as it encounters them, which is all we need to do here.
+        }
     }
 
     readonly #guildBanCache = new Map<string, Set<string>>();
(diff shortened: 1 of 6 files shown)


src/core/BaseUtilities.ts
```diff
@@ -568,17 +577,74 @@ export class BaseUtilities {
...
 
-    readonly #ensuredGuilds = new WeakSet<Guild>();
-    public async ensureMemberCache(guild: Guild): Promise<void> {
-        if (this.#ensuredGuilds.has(guild))
-            return;
+    readonly #guildsWithResolvedMembers = new WeakMap<Guild, Promise<void>>();
+    public ensureMemberCache(guild: Guild): Promise<void> {
+        let resolve = this.#guildsWithResolvedMembers.get(guild);
+        if (resolve === undefined) {
+            this.#guildsWithResolvedMembers.set(guild, resolve = this.#ensureMemberCache(guild).catch(err => {
```
HunteRoi, line 584: Why do you do the `resolve =` bit in here? The parameter is a Promise<void>, right? Wouldn't the rest of the line after the = be enough in this case?

src/core/BaseUtilities.ts
```diff
@@ -568,17 +577,74 @@ export class BaseUtilities {
...
+                throw err;
+            }));
+        }
+        return resolve;
+    }
+
+    async #ensureGuildBans(guild: Guild): Promise<void> {
+        for await (const _ of this.streamAllBans(guild))
+            ;
+    }
```
HunteRoi, line 638: This makes sense as you simply want to cache the guild's bans... But it feels so weird 😆 Also, couldn't the semicolon be placed on the previous line ?

src/core/BaseUtilities.ts
danny-may: Because I need to return the promise that gets created here from this method on line 590. If I dont assign to it then id potentially be returning undefined

src/core/BaseUtilities.ts
danny-may: fair point haha, Ill add a comment to make it clear why we exhaust the bans without using them

